PR #22, opened 2018-01-24
Program total budget estimate in department view

jbothma:


budgetportal/urls.py
```diff
@@ -28,6 +28,9 @@
         '/(?P<government_slug>[\w-]+)'
         '/departments'
         '/(?P<department_slug>[\w-]+).yaml$', cache_page(CACHE_SECS)(views.department)),
+    # Department Budget Totals
```
jbothma: I think this data should be part of the Department view under a key like `programmes - we can probably replace the current `programmes` key with this. We should just ensure that it's compatible.

budgetportal/utils.py
```diff
@@ -0,0 +1,30 @@
+import requests
+from requests.exceptions import HTTPError, ConnectionError
+
+
+def get_budget_resources(year, dept_name):
```
jbothma: I think instead of being a function in a new module, this should be a method of the Department class in `models`.

ebsuku:
OK, latest changes have being pushed.

budgetportal/views.py
```diff
@@ -97,7 +126,7 @@ def department(request, financial_year_id, sphere_slug, government_slug, departm
         'intro': department.intro,
         'treasury_datasets': department.get_treasury_datasets(),
         'contributed_datasets': contributed_datasets if contributed_datasets else None,
-        'programmes': [{'name': p.name} for p in department.programmes.order_by('programme_number')],
+        'programmes': budget,
```
jbothma: How about calling this variable programmes? budget seems vague here

budgetportal/models.py
```diff
@@ -231,6 +233,36 @@ def get_contributed_datasets(self):
...
+            .format(year.slug[:4], self.name),
+            'drilldown': ('activity_programme_number.programme_number|'
+                          'activity_programme_number.programme'),
+            'order': 'activity_programme_number.programme_number:asc',
+            'pagesize': 30
+        }
+        url = ('https://openspending.org/api/3/cubes/'
+               'fb2fa9b200eb3e56facc4c287002fc4d:{}'
+               'aggregate/').format(resource[year.slug[:4]])
+        try:
```
jbothma: I suggest we don't handle this exception and let the programme crash so we don't hide errors which appear like missing data. I tend to work on the principle of only handling exceptions that I really need to - especially when interfacing with other systems - so that I could become aware of unexpected things that can lead to misleading or missing data.

budgetportal/views.py
```diff
@@ -1,7 +1,9 @@
 from django.http import HttpResponse
-from models import FinancialYear, Dataset
+from models import FinancialYear, Dataset, Department
 import yaml
 
+from .utils import get_budget_resources
```
jbothma: We can drop this I think

budgetportal/models.py
```diff
@@ -231,6 +233,36 @@ def get_contributed_datasets(self):
...
+            result = requests.get(url, params=params)
+            if result.status_code == 200:
```
jbothma: you can also use result.raise_for_status() then the exception will tell you what went wrong instead of just ending up with missing data

ebsuku:
Latest changes have being pushed.

jbothma:
Awesome!

I think you can merge and deploy this, then regenerate the data for the static budget portal and merge that if we don't have unexpected weirdness or data loss.

PR #35, opened 2018-02-08
Budget revenue

ebsuku:


budgetportal/views.py
```diff
@@ -4,25 +4,19 @@
...
-            {
-                'category': r['category_two'],
-                'amount': r['amount']
-            }
-        )
 
     context = {
         'selected_financial_year': financial_year_id,
-        'revenue': revenue,
+        'revenue': utils.sort_categories(revenue_data),
```
jbothma: I would tend to avoid "utils" modules - they end up being a bag of catch-all unorganised code. This is something very view-releated - it's a specific view of the revenue data intended for a specific page. I think it fits well in the views module, but if the views module is getting bloated, it could be in a module called something like "summaries" or "aggregates" or something like that

ebsuku:
ready

PR #39, opened 2018-02-13
Department vote match

ebsuku:


budgetportal/models.py
```diff
@@ -298,6 +298,25 @@ def _get_government_query(self):
...
 
+    def _get_primary_department(self):
+        """
+        Check if department is primary
+        """
+        if not self.is_vote_primary:
+            try:
+                name = Department\
+                       .objects \
+                       .get(vote_number=self.vote_number,
```
jbothma: sorry, didn't realise you can filter in `get()` so its MultipleObjectsReturned exception would have been sufficient. But the logging should be helpful. Sorry I made more work.

jbothma:
looks good to me!

PR #40, opened 2018-02-15
add vote primary configs

ebsuku:


budgetportal/models.py
```diff
@@ -314,8 +314,8 @@ def _get_primary_department(self):
                                  "departments" % self.slug)
                 raise
             else:
-                return name.slug
-        return self.slug
+                return name
```
jbothma: watch out, this is a _department_, not a _name_

jbothma:
cool stuff

PR #41, opened 2018-02-16
fixes and utility to add a financial year

jbothma:


budgetportal/models.py
```diff
@@ -172,6 +172,9 @@ def get_function_budgets(self):
             model_url,
             model_result.elapsed.microseconds / 1000
         )
+        if model_result.status_code == 404:
```
jbothma: Perhaps this should actually be allowed to result in exception and be considered a bug

budgetportal/models.py
```diff
@@ -439,6 +443,8 @@ def get_programme_budgets(self):
...
```
jbothma: Perhaps this should actually be allowed to result in exception and be considered a bug

jbothma:
load_departments tested and working at this stage. Didn't try generating the static site with these though.

budgetportal/models.py
jbothma: no this is fine for now - we'll notice very quickly that we're missing this data. And being able to generate the site lets us test things so long

budgetportal/management/commands/load_departments.py
```diff
@@ -0,0 +1,41 @@
+from django.core.management.base import BaseCommand
+from budgetportal.models import Department, Government
+import csv
+from django.utils.text import slugify
+import re
+
+
+class Command(BaseCommand):
+    help = 'load programmes'
```
longhotsummer: copy paste

longhotsummer:
Cool.

PR #47, opened 2018-03-06
Non-financial-year pages

jbothma:


longhotsummer:
This is clever. I hadn't fully grokked that this was how you were building up the yaml. I like the amount of code that this removes.

PR #93, opened 2018-11-01
Add department data slice CSV download URL

otterbotter:
* added functionality to retrieve URL only for aggregate call
* refactored aggregate method in openspending
* added 'department_data_csv' to get_prog_by_econ_budgets response

### DIFF
--- a/budgetportal/csv_gen.py
+++ b/budgetportal/csv_gen.py
@@ -16,6 +16,8 @@ def generate_csv_response(data):
     """
     :param data: dict
     :return: StreamingHttpResponse object
+
+    Roughly based on https://docs.djangoproject.com/en/2.1/howto/outputting-csv/#streaming-large-csv-files
     """
     response = StreamingHttpResponse(streaming_content=iter_items(data['cells'], Echo()),
                                      content_type='text/csv')
@@ -27,9 +29,12 @@ def generate_csv_response(data):
 def iter_items(cells, pseudo_buffer):
     headers_list = cells[0].keys()
     headers_dict = {}
+    writer = csv.DictWriter(pseudo_buffer, fieldnames=sorted(headers_list))
+
+    # Not using writer.writeheader because it's not working - not sure why
     for header in headers_list:
         headers_dict[header] = header
-    writer = csv.DictWriter(pseudo_buffer, fieldnames=sorted(headers_list))
     yield writer.writerow(headers_dict)
+
     for dict_object in cells:
         yield writer.writerow(dict_object)
--- a/budgetportal/csv_gen.py
+++ b/budgetportal/csv_gen.py
@@ -17,19 +17,19 @@ def generate_csv_response(data):
     :param data: dict
     :return: StreamingHttpResponse object
     """
-    response = StreamingHttpResponse(streaming_content=iter_items(data, Echo()),
+    response = StreamingHttpResponse(streaming_content=iter_items(data['cells'], Echo()),
                                      content_type='text/csv')
     response['Content-Disposition'] = 'attachment; filename=\"vulekamali-download.csv\"'
 
     return response
 
 
-def iter_items(data, pseudo_buffer):
-    headers_list = data['cells'][0].keys()
+def iter_items(cells, pseudo_buffer):
+    headers_list = cells[0].keys()
     headers_dict = {}
     for header in headers_list:
         headers_dict[header] = header
     writer = csv.DictWriter(pseudo_buffer, fieldnames=headers_list)
     yield writer.writerow(headers_dict)
-    for dict_object in data['cells']:
+    for dict_object in cells:
         yield writer.writerow(dict_object)
--- a/budgetportal/csv_gen.py
+++ b/budgetportal/csv_gen.py
@@ -19,7 +19,7 @@ def generate_csv_response(data):
     """
     response = StreamingHttpResponse(streaming_content=iter_items(data, Echo()),
                                      content_type='text/csv')
-    response['Content-Disposition'] = 'attachment; filename=\"openspend-dataset.csv\"'
+    response['Content-Disposition'] = 'attachment; filename=\"vulekamali-download.csv\"'
 
     return response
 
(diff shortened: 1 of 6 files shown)


otterbotter:
I believe this is ready to be merged.

budgetportal/csv_gen.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def generate_csv_response(data):
+    """
+    :param data: dict
+    :return: StreamingHttpResponse object
+    """
+    response = StreamingHttpResponse(streaming_content=iter_items(data, Echo()),
+                                     content_type='text/csv')
+    response['Content-Disposition'] = 'attachment; filename=\"openspend-dataset.csv\"'
```
jbothma:
```suggestion
    response['Content-Disposition'] = 'attachment; filename=\"vulekamali-download.csv\"'
```

budgetportal/openspending.py
```diff
@@ -108,18 +135,18 @@ def aggregate(self, cuts=None, drilldowns=None):
...
-        return aggregate_result.json()
+        url = self.cube_url + 'aggregate/'
+        req = requests.Request('GET', url, params=params)
+        prepped_req = self.session.prepare_request(req)
+        prepped_req.params = params  # <- so that we can access params for logging in aggregate()
+        if csv:
+            csv_url = reverse('csv')
+            csv_url += '?api_url=' + urllib.quote(prepped_req.url)
+            if len(csv_url) > URL_LENGTH_LIMIT:
+                raise Exception("Generated URL exceeds %s. Some browsers may no longer be able to interpret the URL.")
```
jbothma: missing argument for string formatting here

budgetportal/views.py
```diff
@@ -320,6 +328,33 @@ def dataset(request, category_slug, dataset_slug):
     return HttpResponse(response_yaml, content_type='text/x-yaml')
 
 
+def json_to_csv(request):
+    """
+    Check if API call is from OpenSpending *
```
jbothma:
```suggestion
    Check if API call is to OpenSpending *
```

budgetportal/views.py
```diff
@@ -320,6 +328,33 @@ def dataset(request, category_slug, dataset_slug):
...
+        return HttpResponse("Invalid domain received: %s (Only openspending.org is allowed)" % domain,
+                            status=403)
+
+    result = requests.get(api_url)
+    logger.info(
+        "request %s took %dms",
+        api_url,
+        result.elapsed.microseconds / 1000
+    )
+    result.raise_for_status()
```
jbothma: Please also add this check here:
```python
        if aggregate_result.json()['total_cell_count'] > PAGE_SIZE:
            raise Exception("More cells than expected - perhaps we should start paging")
```

budgetportal/csv_gen.py
```diff
@@ -0,0 +1,35 @@
...
+
+    return response
+
+
+def iter_items(data, pseudo_buffer):
+    headers_list = data['cells'][0].keys()
```
jbothma: I think make the interface to this module take an argument that is the cell array - so that the caller takes the cell array out of the OpenSpending result - then this module isn't really aware of OpenSpending's response format and that knowledge is a bit more constrained to the json-to-csv view

budgetportal/csv_gen.py
```diff
@@ -0,0 +1,35 @@
...
+    headers_dict = {}
+    for header in headers_list:
+        headers_dict[header] = header
+    writer = csv.DictWriter(pseudo_buffer, fieldnames=headers_list)
```
jbothma: please sort headers_list - that sort order will then be used when you use writeheader()

budgetportal/openspending.py
```diff
@@ -108,18 +135,18 @@ def aggregate(self, cuts=None, drilldowns=None):
...
-        if aggregate_result.json()['total_cell_count'] > PAGE_SIZE:
-            raise Exception("More cells than expected - perhaps we should start paging")
-        return aggregate_result.json()
+        url = self.cube_url + 'aggregate/'
+        req = requests.Request('GET', url, params=params)
+        prepped_req = self.session.prepare_request(req)
+        prepped_req.params = params  # <- so that we can access params for logging in aggregate()
+        if csv:
+            csv_url = reverse('csv')
+            csv_url += '?api_url=' + urllib.quote(prepped_req.url)
```
jbothma: rather make this method always return the OpenSpending API URL:
- drop the params dict logging - we can urldecode the URL if we really wanna see the params
- urlencode the OpenSpending URL and construct the relative URL in the view code - this module method shouldn't be aware of that stuff

budgetportal/csv_gen.py
```diff
@@ -0,0 +1,35 @@
...
+    response = StreamingHttpResponse(streaming_content=iter_items(data, Echo()),
+                                     content_type='text/csv')
+    response['Content-Disposition'] = 'attachment; filename=\"openspend-dataset.csv\"'
+
+    return response
+
+
+def iter_items(data, pseudo_buffer):
+    headers_list = data['cells'][0].keys()
+    headers_dict = {}
```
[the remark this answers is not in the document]
otterbotter: I tried multiple combos - writeheader() never works properly in my experience :(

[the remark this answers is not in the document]
jbothma: cool - this has worked for me in the past but it's not a big deal
```
 with open(csv_file, 'w') as f:
        fieldnames = ['demarcation_code', 'year', 'opinion_code', 'opinion_label']
        writer = csv.DictWriter(f, fieldnames=fieldnames)
        writer.writeheader()
```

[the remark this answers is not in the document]
jbothma: feel free to drop rather than spend much more time on it.

budgetportal/openspending.py
otterbotter: Aggregate() utilises this function to produce it's URL, so if we always return the csv then the other links won't work, or am I missing something?

PR #94, opened 2018-11-02
Caching Programme Budgets

ebsuku:
* On the provincial level, cache all the department programme budgets for a province across all years, then filter by year and department name
* In the national level, cache all department programme budgets for the year, then filter by department name

### DIFF
--- a/budgetportal/models.py
+++ b/budgetportal/models.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from django.core.exceptions import ValidationError, MultipleObjectsReturned
 from django.db import models
 from django.utils.text import slugify as django_slugify
+from django.core.cache import cache
 from itertools import groupby
 from partial_index import PartialIndex
 from pprint import pformat
@@ -487,12 +488,26 @@ class Department(models.Model):
             openspending_api.get_department_name_ref() + ':"' + self.name + '"',
         ]
         if self.government.sphere.slug == 'provincial':
-            cuts.append(openspending_api.get_geo_ref() + ':"%s"' % self.government.name)
+            cuts.append(openspending_api.get_geo_ref() +
+                        ':"%s"' % self.government.name)
+
+            cache_name = 'programme-budgets-provincial-{}'.format(
+                self.government.slug)
+        else:
+            cache_name = 'programme-budgets-national'
         drilldowns = [
             openspending_api.get_programme_number_ref(),
             openspending_api.get_programme_name_ref(),
+            openspending_api.get_department_name_ref()
         ]
-        result = openspending_api.aggregate(cuts=cuts, drilldowns=drilldowns)
+        result_cache = cache.get(cache_name)
+        if result_cache:
+            result = openspending_api.filter_dept(result_cache, self.name,
+                                                  self.government.sphere.slug)
+        else:
+            result = openspending_api.aggregate(
+                cuts=cuts, drilldowns=drilldowns)
+            cache.set(cache_name, result)
         programmes = []
         for cell in result['cells']:
             programmes.append({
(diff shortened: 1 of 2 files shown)


budgetportal/openspending.py
```diff
@@ -121,6 +121,17 @@ def aggregate(self, cuts=None, drilldowns=None):
             raise Exception("More cells than expected - perhaps we should start paging")
         return aggregate_result.json()
 
+    def filter_dept(self, result, dept_name, sphere, financial_year):
+        sphere_filter = {
+            'national': 'vote_number.department',
```
jbothma: I think these names are coincidental - I think if we use self.get_department_dimension() or something like that we can keep writing code using OpenSpending types instead of column names.

budgetportal/openspending.py
```diff
@@ -121,6 +121,17 @@ def aggregate(self, cuts=None, drilldowns=None):
...
+            'provincial': 'department.department'
+        }
+        filtered_results = []
+        for budget in result['cells']:
+            if budget[sphere_filter[sphere]] == dept_name and budget['financial_year.financial_year'] == financial_year:
```
jbothma: here we should also be using self.get_financial_year_ref() or whatever it is instead of 'financia_year.financial_year'

budgetportal/models.py
```diff
@@ -482,17 +483,35 @@ def get_programme_budgets(self):
...
+
         if self.government.sphere.slug == 'provincial':
-            cuts.append(openspending_api.get_geo_ref() + ':"%s"' % self.government.name)
+            cuts.append(openspending_api.get_geo_ref() +
+                        ':"%s"' % self.government.name)
+
+            cache_name = 'programme-budgets-provincial-{}'.format(
+                self.government.slug)
+        else:
+            cache_name = 'programme-budgets-national'
```
jbothma: I think the caching could happen inside openspending.aggregate using the URL as the cache key - then this module doesn't have to be aware of it. What do you think? We'd obviously need to build the full URL, while we currently pass a params dict to requests.get

budgetportal/openspending.py
```diff
@@ -121,16 +120,28 @@ def aggregate_url(self, cuts=None, drilldowns=None):
...
+        else:
+            aggregate_result = self.session.get(url)
+            logger.info("request %s took %dms", aggregate_result.url,
+                        aggregate_result.elapsed.microseconds / 1000)
+            aggregate_result.raise_for_status()
+            aggregate_result = aggregate_result.json()
+            if aggregate_result['total_cell_count'] > PAGE_SIZE:
+                raise Exception(
+                    "More cells than expected - perhaps we should start paging")
+            cache.set(url, aggregate_result)
```
jbothma: can we cache this for 5 minutes?

ebsuku: Default caching is set for 5 min

budgetportal/openspending.py
```diff
@@ -121,16 +120,28 @@ def aggregate_url(self, cuts=None, drilldowns=None):
...
-            aggregate_result.url,
-            aggregate_result.elapsed.microseconds / 1000
-        )
-        aggregate_result.raise_for_status()
-        if aggregate_result.json()['total_cell_count'] > PAGE_SIZE:
-            raise Exception("More cells than expected - perhaps we should start paging")
-        return aggregate_result.json()
+        result_cache = cache.get(url)
+        if result_cache:
+            aggregate_result = result_cache
```
jbothma: please add a cache hit log line

budgetportal/openspending.py
```diff
@@ -121,16 +120,28 @@ def aggregate_url(self, cuts=None, drilldowns=None):
...
+        else:
+            aggregate_result = self.session.get(url)
```
jbothma: please add a cache miss log line

PR #95, opened 2018-11-05
Adjusted budget summary

otterbotter:
Latest addition: direct charge subprogramme changes


budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+
+            ],
+            order=[openspending_api.get_programme_name_ref()])
+        result_for_programmes = openspending_api.filter_dept(result_for_programmes,
+                                                             self.name)
+
+        programme_name_ref = openspending_api.get_programme_name_ref()
+        cells_for_programmes = openspending_api.filter_by_ref_exclusion(result_for_programmes['cells'],
+                                                                        programme_name_ref,
+                                                                        'Direct charge against the National Revenue '
```
jbothma: We can factor this out to a constant

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+            'direct_charges': self._get_budget_direct_charges(openspending_api),
+            'department_data_csv': csv_url(dept_aggregate_url),
+        }
+
+    def _get_adjustments_by_type(self, openspending_api, cells):
+        budget_phase_ref = openspending_api.get_phase_ref()
+        adjustment_kind_ref = openspending_api.get_adjustment_kind_ref()
+
+        def filter_by_type(cell):
+            types = [
```
jbothma: We could factor this out to a constant at the top of the module

README.md
```diff
@@ -45,6 +45,44 @@ Then run the server
 python manage.py runserver_plus
 ```
 
+### Deploy to staging
```
jbothma: I think these changes fit better in the Manual - I think the README of this app should focus on this app, and not so much where it's deployed. This is especially the case with the Static and Teplates section - that stuff isn't part of this app, so I think it's up to the Maintenance Manual's development section to bring this app and static-budget-portal together. Do you think it's actually really important to have it here?

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+            params = {
+                'sql': sql
+            }
+            result = requests.get(CKAN_DATASTORE_URL, params=params)
+            result.raise_for_status()
+            records = result.json()['result']['records']
+            if records:
+                value = records[0]['Value of Virements']
+            else:
+                raise Exception('Value of Virements query returned no records')
```
jbothma: you could even have let this just be an index out of range error and skipped the condition, since we're not handling it in a special way - the exception would lead us right to the problem.

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+                drilldowns=openspending_api.get_all_drilldowns())
+
+            result_for_virements = openspending_api.filter_dept(result_for_virements, self.name)
+            cells_for_virements = openspending_api.filter_by_ref_exclusion(result_for_virements['cells'],
+                                                                           openspending_api.get_programme_name_ref(),
+                                                                           'Direct charge against the National '
+                                                                           'Revenue Fund')
+            total_positive_virement_change = 0
+            for c in cells_for_virements:
+                value = c['value.sum']
```
jbothma: it's cool to do assignment like this for readability. Don't worry about doing it for efficiency - you won't see the benefit of factoring out a dict lookup except in very special cases.

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+                openspending_api.get_department_name_ref(),
+                openspending_api.get_adjustment_kind_ref(),
+            ],
+            order=[openspending_api.get_subprogramme_name_ref()])
+        result_for_direct_charges = openspending_api.filter_dept(result_for_direct_charges, self.name)
+
+        subprog_ref = openspending_api.get_subprogramme_name_ref()
+        phase_ref = openspending_api.get_phase_ref()
+        kind_ref = openspending_api.get_adjustment_kind_ref()
+        subprog_dict = {}
```
jbothma: since we stick them into a dict we lose ordering - we should ensure consistent ordering - one option is to use OrderedDict here. Another is to sort later.

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+        for cell in result_for_direct_charges['cells']:
+            if cell[kind_ref] == 'Adjustments - Total adjustments':
+                subprog_dict[cell[subprog_ref]] = {
+                    'amount': cell['value.sum'],
+                    'label': cell[subprog_ref]
+                }
+
+        for subprog in subprog_dict.keys():
+            for cell in result_for_direct_charges['cells']:
+                if cell[subprog_ref] == subprog:
```
jbothma: if we write a function like `openspending.filter(cells, list_of_ref_value_tuples) -> matching_cells` we could factor out this kind of code a lot.


It could take an optional argument expected_cell_count which raises an exception if the cell count doesn't match the result.

budgetportal/models.py
```diff
@@ -829,6 +1152,7 @@ def __init__(self, **kwargs):
         self.category = kwargs['category']
         self._openspending_api = None
         self.package = kwargs['package']
+        self._adjusted_budget_summary = {}
```
jbothma: I think this was added by mistake. This caching on `Department` is also me being silly and we can clean that up.

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+                openspending_api.get_adjustment_kind_ref(),
+                openspending_api.get_phase_ref(),
+                openspending_api.get_programme_name_ref(),
+                openspending_api.get_department_name_ref()
+            ],
+            order=[openspending_api.get_adjustment_kind_ref()])
+
+        result_for_type_and_total = openspending_api.filter_dept(result_for_type_and_total,
+                                                                 self.name)
+        cells_for_type_and_total = openspending_api.filter_and_aggregate(
```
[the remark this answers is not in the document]
otterbotter: You're right - I've implemented this to make it easier to change the logic since we are using filter and aggr in combination in many places, but it is not intuitive and not likely to be used again in the future. Thanks.

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+                                                                       [openspending_api.get_econ_class_2_ref(),
+                                                                        openspending_api.get_econ_class_3_ref()])
+
+        for cell in cells_for_econ_classes:
+            new_econ_2_object = {'type': 'economic_classification_3',
+                                 'name': cell[econ_class_3_ref],
+                                 'amount': cell['value.sum']}
+            if cell[econ_class_2_ref] not in econ_classes.keys():
+                if cell['value.sum'] != 0:
+                    econ_classes[cell[econ_class_2_ref]] = dict()
```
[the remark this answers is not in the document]
otterbotter: Thanks for pointing that out

budgetportal/models.py
```diff
@@ -768,8 +769,324 @@ def get_expenditure_over_time(self):
...
+                    econ_classes[cell[econ_class_2_ref]]['type'] = 'economic_classification_2'
+                    econ_classes[cell[econ_class_2_ref]]['name'] = cell[econ_class_2_ref]
+                    econ_classes[cell[econ_class_2_ref]]['items'] = [new_econ_2_object]
+            else:
```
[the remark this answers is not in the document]
otterbotter: Much better, thanks!

budgetportal/models.py
otterbotter: Good point, thanks.

budgetportal/models.py
otterbotter: It was definitely for readability, this is a programming habit of mine. Though I think in this case it's unnecessary as it's only used for one conditional and one operation

budgetportal/models.py
otterbotter: thanks, did not consider ordering. i've implemented an ordereddict here as suggested

budgetportal/models.py
otterbotter: yep, shouldn't be here - thanks for spotting that.

budgetportal/models.py
otterbotter: I don't see many repetitions of this type of code in the section we've been working on recently.
Could I look at this again when I need to do more similar operations?

PR #99, opened 2018-11-23
Department name update fix (Treasury datasets removal)

otterbotter:


budgetportal/models.py
```diff
@@ -228,8 +230,6 @@ class Meta:
         ordering = ['vote_number', 'name']
 
     def save(self, *args, **kwargs):
```
jbothma: I'm sure you can totally remove this method now

otterbotter: good point thanks - i've removed it

PR #105, opened 2018-12-12
budget vs actual using annual report data

jbothma:


budgetportal/models.py
```diff
@@ -46,6 +48,16 @@
     '2018-19': '5b315ff0-55e9-4ba8-b88c-2d70093bfe9d',
 }
 
+FISCAL_YEAR_MAPPING = {
```
jbothma: you can use `FinancialYear.slug_from_year_start` for this

PR #112, opened 2019-01-29
Add Department website link

otterbotter:
⚠️ **This PR is not ready to be merged and is still in draft form.**

This PR:
- Adds a website link for each department
- Always retrieves the latest department link (even if it is null)
- Adds **x** new helper functions
-- list them here
- Adds the website url to the import and export functions
- Adds automated tests (in progress)
- Adds a csv export view for departments (similar to the programme csv export)


budgetportal/management/commands/load_departments.py
```diff
@@ -34,39 +34,58 @@ def add_arguments(self, parser):
...
-                if row.get('vision', False):
-                    intro += "## Vision\n\n%s\n\n" % row['vision']
-                if row.get('mission', False):
-                    intro += "## Mission\n\n%s\n\n" % row['mission']
-                if row.get('mandate', False):
-                    intro += "## Mandate\n\n%s\n\n" % row['mandate']
-                if row.get('core functions and responsibilities', False):
-                    intro += "## Core functions and responsibilities\n\n%s\n\n" % row['core_functions_and_responsibilities']
+                website_url = None
+                if row.get('intro', False):
```
jbothma: just assign here - don't see any need for append any more?

I think you should make the default None rather than False - then the possible types here are (string or None) rather than (string or boolean)

budgetportal/models.py
```diff
@@ -288,6 +297,29 @@ def get_govt_functions(self):
...
+        latest_existing_department = None
+        found_or_out_of_years, index = False, 0
+        while not found_or_out_of_years:
+            year = years[index]
+            index += 1
+            if year == original_dept_year:
+                found_or_out_of_years = True
+
+            latest_sphere = Sphere.objects.get(financial_year=year, name=sphere_name)
+            latest_government = latest_sphere.governments.get(name=self.government.name)
```
jbothma: if there's nothing else needing this intermediate functionality, wouldn't it be better to do something like
```python
Department.objects.filter(government__slug=self.government.slug,
                                        government__sphere__slug=self.government.sphere.slug,
...
)
```
?

Then we can avoid all this logic til it's needed

budgetportal/tests/test_data/test_management_commands_national_departments.csv
```diff
@@ -0,0 +1,41 @@
+government,department_name,vote_number,is_vote_primary,intro,website_url
```
jbothma: could you use more minimal test data so it's a bit easier to verify what's important? Doesn't have to be super realistic

budgetportal/tests/test_data/test_management_commands_provincial_departments.csv
```diff
@@ -0,0 +1,70 @@
+government,department_name,vote_number,is_vote_primary,intro,website_url
```
jbothma: ditto

budgetportal/tests/test_management_commands.py
```diff
@@ -21,52 +21,52 @@ def setUp(self):
 
         # governments
         self.south_africa = Government.objects.create(sphere=national, name='South Africa')
-        self.fake_central_province = Government.objects.create(
+        self.free_state = Government.objects.create(
```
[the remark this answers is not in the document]
otterbotter: ok - good idea. i've left the data itself as real data, but i've renamed the variables - not sure if this defeats the purpose

let me know if you want me to alter the test data to look more fake

budgetportal/tests/test_department.py
```diff
@@ -211,3 +211,41 @@ def test_missing_data_dept_did_not_exist(self):
         self.mock_openspending_api.aggregate_by_three_ref = Mock(return_value=self.mock_data['test_cells_data_missing_2018'])
         result = self.department.get_expenditure_time_series_summary()
         self.assertEqual(result['notices'], ['This department did not exist for some years displayed.'])
+
+
+class DepartmentWebsiteUrlTestCase(TestCase):
+    """ Unit tests of budgeted and actual expenditure summary for a department """
```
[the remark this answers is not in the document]
otterbotter: i see what you're referring to now: the comment says unit test when it's not. good catch - thanks.

budgetportal/models.py
otterbotter: Changed to your solution and had to add `slug=self.slug` to prevent other departments from being returned - thank you.

budgetportal/tests/test_data/test_management_commands_national_departments.csv
otterbotter: as discussed on Slack, will reduce the amount of data in the `intro` field

PR #118, opened 2019-02-15
Infrastructure Projects

otterbotter:
This PR's primary function is to introduce the Infrastructure Projects section to vulekamali.

➕ New additions:
- `InfrastructureProject` class
- Infrastructure Project automated unit tests
- Infrastructure Projects index page `/infrastructure-projects.yaml`
- Infrastructure Projects detail page `/infrastructure-projects/project-slug-here.yaml`

🔨 Changes:
- CKAN host url in models now uses env variable's root domain to support development with a local CKAN instance

📓 General notes:
- YAML output is always sorted to prevent unnecessary git diffs
- Only unit tests were written, no integration tests. I'd like to chat to @jbothma about the best way to inject data into the system, to maximize the extent of the integration test.

### DIFF
--- a/budgetportal/models.py
+++ b/budgetportal/models.py
@@ -289,7 +289,7 @@ class Department(models.Model):
 
     def get_website_url(self):
         """ Always return the latest available URL, even for old departments. """
-        return self._get_latest_department_instance().website_url
+        return self.get_latest_department_instance().website_url
 
     def get_url_path(self):
         return "%s/departments/%s" % (self.government.get_url_path(), self.slug)
@@ -300,7 +300,7 @@ class Department(models.Model):
     def get_financial_year(self):
         return self.government.sphere.financial_year
 
-    def _get_latest_department_instance(self):
+    def get_latest_department_instance(self):
         """ Try to find the department in the most recent year with the same slug.
         Continue traversing backwards in time until found, or until the original year has been reached. """
         newer_departments = Department.objects.filter(government__slug=self.government.slug,
--- a/budgetportal/models.py
+++ b/budgetportal/models.py
@@ -33,7 +33,7 @@ ckan = settings.CKAN
 
 URL_LENGTH_LIMIT = 2000
 
-CKAN_DATASTORE_URL = ('https://data.vulekamali.gov.za'
+CKAN_DATASTORE_URL = (settings.CKAN_URL+
                       '/api/3/action' \
                       '/datastore_search_sql')
 
@@ -1432,6 +1432,67 @@ class InfrastructureProject():
         self.coordinates = []
         self.provinces = []
         self.expenditure = []
+        self.dataset = None
+
+    @classmethod
+    def get_dataset(cls):
+        """ Return the first dataset in the Infrastructure Projects group. """
+        query = {
+            'q': '',
+            'fq': ('+organization:"national-treasury"'
+                   '+vocab_spheres:"national"'
+                   '+groups:"infrastructure-projects"'),
+            'rows': 1,
+        }
+        response = ckan.action.package_search(**query)
+        logger.info(
+            "query %s\nreturned %d results",
+            pformat(query),
+            len(response['results'])
+        )
+        if response['results']:
+            return Dataset.from_package(response['results'][0])
+        else:
+            raise Exception('Could not find Infrastructure Projects dataset')
+
+    @classmethod
+    def get_project_from_resource(cls, project_slug):
+        """ Uses first CSV resource in dataset """
+        resource = cls.get_dataset().get_resource(format='CSV')
+        sql = '''
+                SELECT * FROM "{}" WHERE "Featured"='TRUE' AND "Project slug"='{}' 
+            '''.format(resource['id'], project_slug)
+        params = {'sql': sql}
+        revenue_result = requests.get(CKAN_DATASTORE_URL, params=params)
+        revenue_result.raise_for_status()
+        revenue_data = revenue_result.json()['result']['records']
+        project = InfrastructureProject(records=revenue_data)
+        return project
+
+    @classmethod
+    def get_featured_projects_from_resource(cls):
+        """ Uses first CSV resource in dataset """
+        resource = cls.get_dataset().get_resource(format='CSV')
+        sql = '''
+                SELECT * FROM "{}" WHERE "Featured"='TRUE'
+            '''.format(resource['id'])
+
+        params = {'sql': sql}
+        projects_result = requests.get(CKAN_DATASTORE_URL, params=params)
+        projects_result.raise_for_status()
+        revenue_data = projects_result.json()['result']['records']
+
+        # Assume project names are unique within the subset of featured projects
+        unique_project_names = []
+        for obj in revenue_data:
+            if obj['Project name'] not in unique_project_names:
+                unique_project_names.append(obj['Project name'])
+
+        projects = []
+        for project_name in unique_project_names:
+            project_list = filter(lambda x: x['Project name'] == project_name, revenue_data)
+            projects.append(InfrastructureProject(records=project_list))
+        return projects
 
     def get_url_path(self):
         return "/infrastructure-projects/{}".format(slugify(self.name))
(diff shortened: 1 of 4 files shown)


budgetportal/models.py
```diff
@@ -1412,6 +1416,148 @@ class PackageWithoutGroupException(Exception):
...
+                   '+groups:"infrastructure-projects"'),
+            'rows': 1,
+        }
+        response = ckan.action.package_search(**query)
+        logger.info(
+            "query %s\nreturned %d results",
+            pformat(query),
+            len(response['results'])
+        )
+        if response['results']:
```
jbothma: looks like this will crash in production until we upload this data? But we need to be able to deploy this and just have the data appear in the static site once it gets uploaded to ckan

budgetportal/models.py
```diff
@@ -1412,6 +1416,148 @@ class PackageWithoutGroupException(Exception):
     pass
 
 
+class InfrastructureProject():
+    """ Represents an infrastructure project stored in CKAN """
+
+    def __init__(self, **kwargs):
+        self.records = kwargs.get('records')
+        if self.records:
```
jbothma: when would you want an InfrastructureProject instance that isn't initialised with data? I think you'd normally want to be able to assume it's initialised or instantiation should fail

budgetportal/models.py
```diff
@@ -1412,6 +1416,148 @@ class PackageWithoutGroupException(Exception):
...
+            project_list = filter(lambda x: x['Project name'] == project_name, revenue_data)
+            projects.append(InfrastructureProject(records=project_list))
+        return projects
+
+    def get_url_path(self):
+        return "/infrastructure-projects/{}".format(slugify(self.department_name + '-' + self.name))
+
+    def get_projected_expenditure(self):
+        # Include only projected amounts
+        projected_project_list = filter(lambda x: int(x['Financial Year']) > 2018, self.records)
```
jbothma: we can't hardcode this - you'll have to go by budget_phase - projected expenditure is the MTEF years (medium term economic framework)

budgetportal/models.py
```diff
@@ -1412,6 +1416,148 @@ class PackageWithoutGroupException(Exception):
...
+            if 'and' in self.gps_codes:
+                gps_codes_grouped = self.gps_codes.split('and')
+                for code_group in gps_codes_grouped:
+                    lat_long = [float(x) for x in code_group.split(',')]
+                    self.coordinates.append({
+                        'latitude': lat_long[0],
+                        'longitude': lat_long[1]
+                    })
+            elif ',' in self.gps_codes:
+                lat_long = [float(x) for x in self.gps_codes.split(',')]
```
jbothma: this repeated bit could be split out as `_parse_coordinate` which takes a string and returns a `{latitude: ..., longitude: ...}`

then it's also easy to test that it does what you want

budgetportal/models.py
```diff
@@ -1412,6 +1416,148 @@ class PackageWithoutGroupException(Exception):
...
+    def get_projected_expenditure(self):
+        # Include only projected amounts
+        projected_project_list = filter(lambda x: int(x['Financial Year']) > 2018, self.records)
+
+        projected_expenditure = 0
+        for project in projected_project_list:
+            projected_expenditure += float(project['Amount'])
+        return projected_expenditure
+
+    def get_cleaned_coordinates(self):
```
jbothma: this whole thing could be functional, and called in the constructor like get_cleaned_coordinates(gps_codes) then it's really easy to test

budgetportal/models.py
```diff
@@ -1412,6 +1416,148 @@ class PackageWithoutGroupException(Exception):
...
+                               format(self.name, self.gps_codes))
+        except Exception as e:
+            logger.warning("Caught Exception '{}' for co-ordinates for infrastructure project '{}'".format(
+                e, self.name
+            ))
+        return self.coordinates
+
+    def get_provinces(self):
+        params = {'type': 'PR'}
+        if not self.coordinates:
```
jbothma: if you parse coordinates in the constructor, you don't need to think about whether they're cleaned in this method

budgetportal/models.py
```diff
@@ -1412,6 +1416,148 @@ class PackageWithoutGroupException(Exception):
...
+            self.get_cleaned_coordinates()
+        for c in self.coordinates:
+            province_result = requests.get(MAPIT_POINT_API_URL.format(c['longitude'], c['latitude']), params=params)
+            province_result.raise_for_status()
+            list_of_objects_returned = province_result.json().values()
+            if len(list_of_objects_returned) > 0:
+                province_name = list_of_objects_returned[0]['name']
+                if province_name not in self.provinces:
```
jbothma: you could use a [set](https://docs.python.org/2/library/sets.html) as an intermediate datastructure, and just turn it into a list when this method is done so you don't need to do `if province_name not in ...`

budgetportal/tests/test_infrastructure_projects.py
```diff
@@ -0,0 +1,100 @@
...
+            {'Financial Year': 2017, 'Amount': 100},
+            {'Financial Year': 2018, 'Amount': 100},
+            {'Financial Year': 2019, 'Amount': 100},
+            {'Financial Year': 2020, 'Amount': 100},
+            {'Financial Year': 2021, 'Amount': 100}
+        ]
+        self.project = InfrastructureProject()
+
+    def test_success(self):
+        self.project.records = self.fake_records
```
jbothma: this is a bit dodgy - is _setting_ `.records` intended to be a public interface? I reckon you might actually never want to set `InfraProject.records`, or at least mark it private as `_records`

you don't want to modify the class directly - you should instantiate it with the test data. it's not so much the test that's not right, as the fact that all the instantiation doesn't happen in the constructor.

An alternative is to extend the class with a class with the test data hardcoded and just used by `get_projected_expenditure` but that's very white-box.

budgetportal/tests/test_infrastructure_projects.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class CleanedCoordinatesTestCase(TestCase):
+    """ Unit tests for get_cleaned_coordinates function """
+
+    def setUp(self):
+        self.project = InfrastructureProject()
+
+    def test_standard_format(self):
+        self.project.gps_codes = '-26.378582,27.654933'
```
jbothma: like I mentioned above, if you factor out the coordinate parsing to one or two functions, you can test those directly as functional units without needing to instantiate InfrastructureProject() each time

budgetportal/models.py
```diff
@@ -1412,9 +1417,178 @@ class PackageWithoutGroupException(Exception):
...
+        response = ckan.action.package_search(**query)
+        logger.info(
+            "query %s\nreturned %d results",
+            pformat(query),
+            len(response['results'])
+        )
+        if response['results']:
+            return Dataset.from_package(response['results'][0])
+        else:
+            raise Http404()
```
jbothma: This works, although it'd be better if models.py indicated that it's missing to the view code, and the view code itself raised the 404, since a 404 is very much an http thing which is the view's concern, while the model is concerned with data and not the protocol used to access it.

budgetportal/models.py
```diff
@@ -1412,9 +1417,178 @@ class PackageWithoutGroupException(Exception):
...
+        """ Uses first CSV resource in dataset """
+        resource = cls.get_dataset().get_resource(format='CSV')
+        sql = '''
+                SELECT * FROM "{}" WHERE "Featured"='TRUE'
+            '''.format(resource['id'])
+
+        params = {'sql': sql}
+        projects_result = requests.get(CKAN_DATASTORE_URL, params=params)
+        projects_result.raise_for_status()
+        revenue_data = projects_result.json()['result']['records']
```
jbothma:
```suggestion
        project_records = projects_result.json()['result']['records']
```

budgetportal/models.py
```diff
@@ -1412,9 +1417,178 @@ class PackageWithoutGroupException(Exception):
...
+
+        # Assume project names are unique within the subset of featured projects
+        unique_project_names = []
+        for obj in revenue_data:
```
jbothma:
```suggestion
        for obj in project_records:
```

PR #123, opened 2019-02-21
Fix website_url null for 2019

otterbotter:
Code can be optimized to chain filter inside conditional - but could not get it working locally yet for some reason. This will do to at least get the urls up for now.

### DIFF
--- a/budgetportal/models.py
+++ b/budgetportal/models.py
@@ -296,7 +296,7 @@ class Department(models.Model):
 
     def get_website_url(self):
         """ Always return the latest available URL, even for old departments. """
-        return self.get_latest_department_instance().website_url
+        return self.get_latest_department_instance(require_website_url=True).website_url
 
     def get_url_path(self):
         return "%s/departments/%s" % (self.government.get_url_path(), self.slug)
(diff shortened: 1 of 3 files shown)


budgetportal/models.py
```diff
@@ -307,12 +307,19 @@ def get_govt_functions(self):
...
         return self.government.sphere.financial_year
 
-    def get_latest_department_instance(self):
+    def get_latest_department_instance(self, require_website_url=False):
         """ Try to find the department in the most recent year with the same slug.
         Continue traversing backwards in time until found, or until the original year has been reached. """
-        newer_departments = Department.objects.filter(government__slug=self.government.slug,
-                                                      government__sphere__slug=self.government.sphere.slug,
-                                                      slug=self.slug).order_by('-government__sphere__financial_year__slug')
+        if require_website_url:
```
jbothma: I think this should rather be two totally different methods:

`Department.get_latest_department_instance` and `Department.get_latest_department_url`

It doesn't really make sense to have a modifier argument to `Department.get_latest_department_instance` for an arbitrary field

jbothma: the code will then be simpler and more self-explanatory

PR #124, opened 2019-03-11
Pdf document link fix for infrastructure projects

otterbotter:
This PR:
- Fixes a bug where the `pdf_document_url` link on the `Infrastructure Project` pages was always null in production
- Refactors the code to retrieve `pdf_document_url` into a function on the `InfrastructureProject` class

### DIFF
--- a/budgetportal/models.py
+++ b/budgetportal/models.py
@@ -1513,6 +1513,26 @@ class InfrastructureProject:
     def get_url_path(self):
         return "/infrastructure-projects/{}".format(self.slug)
 
+    def get_budget_document_url(self, document_format='PDF'):
+        """
+        Returns budget-vote-document URL, for given format,
+        if the latest department instance matches the project year
+        """
+        departments = Department.objects.filter(
+            slug=slugify(self.department_name),
+            government__sphere__slug='national'
+        )
+        if departments:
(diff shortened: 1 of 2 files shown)


budgetportal/models.py
```diff
@@ -1513,6 +1513,26 @@ def get_featured_projects_from_resource(cls):
     def get_url_path(self):
         return "/infrastructure-projects/{}".format(self.slug)
 
+    def get_budget_document_url(self, document_format='PDF'):
+        """
+        Returns budget-vote-document URL, for given format,
+        if the latest department instance matches the project year
```
jbothma:
```suggestion
        if the latest department instance matches the project dataset year
```

budgetportal/models.py
```diff
@@ -1513,6 +1513,26 @@ def get_featured_projects_from_resource(cls):
...
+        """
+        departments = Department.objects.filter(
+            slug=slugify(self.department_name),
+            government__sphere__slug='national'
+        )
+        if departments:
+            latest_dept = departments[0].get_latest_department_instance()
+            project_year = InfrastructureProject.get_dataset().package['financial_year'][0]
```
jbothma: how about `project_data_year`?

PR #127, opened 2019-03-29
Provincial treemap data generation endpoints

otterbotter:


budgetportal/openspending.py
```diff
@@ -159,12 +159,18 @@ def get_subprogramme_dimension(self):
...
+        return self.get_ref(self.get_government_dimension(), 'key')
+
     def get_vote_number_ref(self):
         return self.get_ref(self.get_department_dimension(), 'key')
 
     def get_department_dimension(self):
         return self.get_dimension('administrative_classification')
 
+    def get_government_dimension(self):
+        return self.get_dimension('administrative_classification', level=1)
```
jbothma: this isn't right - government should be geo which exists - I think when you uploaded provincial budget vs actual you set government to administrative when it should be geo source and now when you query for it you have to use administration.

administration is used for department

budgetportal/views.py
```diff
@@ -21,16 +21,20 @@
...
+        context = dept.get_national_expenditure_treemap(financial_year_id, phase_slug)
+        response_yaml = yaml.safe_dump(context, default_flow_style=False, encoding='utf-8')
+        return HttpResponse(response_yaml, content_type='text/x-yaml')
+    elif sphere_slug == 'provincial':
+        dept = Department.objects.filter(government__sphere__slug='provincial')[0]
+        context = dept.get_provincial_expenditure_treemap(financial_year_id, phase_slug)
+        response_yaml = yaml.safe_dump(context, default_flow_style=False, encoding='utf-8')
+        return HttpResponse(response_yaml, content_type='text/x-yaml')
+    else:
+        return HttpResponse("Unknown government sphere. Options are: national", status=400)
```
jbothma: should remove or fix the options bit - it's currently missing provincial.

budgetportal/openspending.py
otterbotter: ah okay, makes sense. I wasn't sure which dimension fit best. In this case department is top level and gov is level 1.

PR #147, opened 2019-05-15
Load departments using GUI instead of django manage command

delenamalan:


delenamalan:
Example error message:
Screenshot at 2019-05-16 00-48-15

delenamalan:
I see the "department name" and "government" column headers are switched around in the screenshot above. It's ordered correctly though when the data set doesn't cause any errors:
Screenshot at 2019-05-16 00-58-32

PiDelport:
Card: 
https://trello.com/c/O3xFtQJ2/26-load-departments-using-gui-instead-of-django-manage-command

PiDelport:
Review in progress: I'm wrangling some PhantomJS things.

budgetportal/import_export_admin.py
```diff
@@ -0,0 +1,148 @@
...
+        unique government-vote_number combination.
+        """
+        name = self.resource.fields['name'].clean(row)
+        slug = slugify(name)
+        government = self.resource.fields['government'].clean(row)
+
+        instance = None
+
+        try:
+            instance = Department.objects.all().get(name=name, government=government)
```
jbothma: why `.all().get()` and not just `.get()`?

Also a couple more of these below

delenamalan: Sorry, don't know why I did that. Fixing it now.

budgetportal/tests/test_admin_departments_upload.py
```diff
@@ -0,0 +1,126 @@
...
+
+        super(AdminDepartmentUploadTest, self).setUp()
+
+    def tearDown(self):
+        self.selenium.quit()
+
+        # https://stackoverflow.com/questions/14991244/how-do-i-capture-a-screenshot-if-my-nosetests-fail
+        if sys.exc_info()[0]:  # Returns the info of exception being handled
+            now = datetime.now().strftime('%Y-%m-%d_%H-%M-%S-%f')
+            self.selenium.get_screenshot_as_file('%s.png' % now)
```
PiDelport: This screenshot code fails for me, with the WebDriver code unable to connect to the server to send the screenshot command. I think it only actually works before `selenium.quit()`, not after.

(If it's working sometimes, it could be due to a race condition between the quit and screenshot commands?)

PiDelport: Should be fixed in 4f8ccd2.

budgetportal/import_export_admin.py
```diff
@@ -0,0 +1,148 @@
...
+        queryset=Sphere.objects.all(),
+        required=True
+    )
+
+
+class CustomCSV(base_formats.TextFormat):
+    """
+    Class that will be used in the django-import-export module
+    to import data from a csv file. We override the create_dataset method
+    because the package's create_dataset method doesn't seem to work.
```
PiDelport: This is due to https://github.com/django-import-export/django-import-export/issues/957 (tablib 0.13+ breaks django-import-export 1.2.0)

We can avoid the need for this workaround by keeping tablib to 0.12, which should be simpler and better overall. (The next release of django-import-export should fix this, but will also officially drop Python 2 support.)

PiDelport: c81e1da should address this.

budgetportal/import_export_admin.py
```diff
@@ -0,0 +1,135 @@
...
+            try:
+                instance = Department.objects.get(
+                    vote_number=vote_number, government=government)
+            except Department.DoesNotExist:
+                pass
+
+        if instance:
+            return instance
+
+        return None
```
PiDelport: We could probably simplify this method body significantly… something like this?

```python
        q = Q(name=name, government=government)
        q |= Q(slug=slug, government=government)
        if is_vote_primary:
            q |= Q(vote_number=vote_number, government=government)

        return Department.objects.filter(q).first()
```

delenamalan: Oh nice! Will use that!

PR #149, opened 2019-05-17
Use latest cpi dataset automatically

delenamalan:


PiDelport:
Card: https://trello.com/c/CPKQZ3Px/64-use-latest-cpi-dataset-automatically

jbothma:
No changes to data when building against this in staging.

budgetportal/tests/test_models.py
```diff
@@ -0,0 +1,57 @@
+from django.conf import settings
```
jbothma: this suite should be called test_datasets now, rather than test_models

delenamalan:
@jbothma Sounds right :+1:

PR #151, opened 2019-05-20
Treat an empty function column as if function data has not been uploaded

jbothma:
https://trello.com/c/IykAdd9F/240-treat-blank-only-function-in-provincial-data-as-not-uploaded-in-focus-area-pages


budgetportal/tests/test_pages.py
```diff
@@ -9,6 +9,11 @@
 from mock import Mock
 import yaml
 
+# Hacky make sure we don't call out to openspending.
+import requests
+requests.get = Mock
+requests.Session = Mock
```
PiDelport: We could use the https://github.com/getsentry/responses library here?

PiDelport: (And for the other occurrences.)

PR #156, opened 2019-07-26
Migrate jekyll to django - Step 1 - /about

havanhuy1997:
I remove unused files, please check


jbothma:
To check that there are no differences between directory trees that should be copied verbatim:

```
(jdb)-(12:25 PM Mon Jul 29)-(~/proj/code4sa/treasury-portal/datamanager):
 (havanhuy1997-migrate-juekyll-to-django) $ colordiff -x node_modules -r ../static-budget-portal/packages/ packages/ | less -R
```

jbothma:
Needs documentation of yarn install, build...
This could be done in a later task since this was more of a tracer bullet.

jbothma:
Nice job

```diff
diff -r ../static-budget-portal/_includes/components/header-and-footer/ExtraFooter/index.html _includes/components/header-and-footer/ExtraFooter/index.html
1a2
> {% load static %}
7c8
<         <img width="200" class="ExtraFooter-image ExtraFooter-image--larger" src="/assets/images/treasury-logo.jpg"/>
---
>         <img width="200" class="ExtraFooter-image ExtraFooter-image--larger" src="{% static 'images/treasury-logo.jpg' %}"/>
10c11
<         <img width="200" class="ExtraFooter-image" src="/assets/images/imali-yethu-logo-h.png"/>
---
>         <img width="200" class="ExtraFooter-image" src="{% static 'images/imali-yethu-logo-h.png' %}"/>
```

jbothma:
@havanhuy1997 There's a javascript error on the about page that isn't on the current about page - Why is that? Is that something that's been forgotten or how can it be fixed?

Screenshot_2019-07-29_14-31-10

havanhuy1997:
Okay, I will check again. I didn't see that error before

havanhuy1997:
Screenshot from 2019-07-29 15-41-46
I see the original page has similar errors

havanhuy1997:
i will debug and see what this happens

jbothma:
Departments link isn't generated correctly (I know it won't link to a working page, but currently it's not even a valid url)

```diff
409c421
<             <a class="NavBar-link" href="https://vulekamali.gov.za/2019-20/departments">
---
>             <a class="NavBar-link" href="http://departments/">
```

jbothma:
Background in-page-nav item has undexpected encoding - that's probably what makes it not show as selected like the prod one does

```diff
576c643
<         <a class="SubLinks-subLink js-link is-active" data-connected-id="background" href="https://vulekamali.gov.za/about#background">
---
>         <a class="SubLinks-subLink js-link" data-connected-id="&quot;background&quot;" href="http://localhost:8000/about#background">
```

jbothma:
Querystring value of this seems to be wrong
```diff
<       <a class="Button is-secondary is-inline" target="_blank" href="https://docs.google.com/forms/d/e/1FAIpQLSdTgVXAPqmiT33VF_GvMtEANhxIi3V9BlVrqsHoI0Xy1Bv14w/viewform?entry.1530467677=https%3A%2F%2Fvulekamali.gov.za%2F%2F2018-19%2Ffocus%2F0.html">
---
>       <a class="Button is-secondary is-inline" target="_blank" href="https://docs.google.com/forms/d/e/1FAIpQLSdTgVXAPqmiT33VF_GvMtEANhxIi3V9BlVrqsHoI0Xy1Bv14w/viewform?entry.1530467677=https://vulekamali.gov.za/">
```

jbothma:
We've lost the Google Tag Manager snippets - is that because they were disabled by jekyll environment?

It should probably be disabled by django stettings.DEBUG now - if that's already the case please link to how that's done.

```diff
57,64d56
<   <!-- Google Tag Manager -->
<   <script async="" src="About%20-%20vulekamali-prod_files/gtm.js"></script><script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
<   new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
<   j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
<   'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
<   })(window,document,'script','dataLayer','GTM-PLNL56L');</script>
<   <!-- End Google Tag Manager -->
...
<     <!-- Google Tag Manager (noscript) -->
<     <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-PLNL56L"
<                       height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
<     <!-- End Google Tag Manager (noscript) -->
```

package.json
```diff
@@ -0,0 +1,134 @@
...
+    "packages/webapp"
+  ],
+  "nohoist": [
+    "packages/webapp"
+  ],
+  "name": "static-budget-portal",
+  "version": "1.0.0",
+  "scripts": {
+    "build:webapp": "cd ./packages/webapp && node ./scripts/build.js",
+    "build:root": "webpack --display errors-only --progress -p --watch",
```
jbothma: `--watch` should be on buid:dev not build:root

jbothma:
> Screenshot from 2019-07-29 15-41-46
> I see the original page has similar errors

I'm talking about `e is null`

.gitignore
```diff
@@ -5,4 +5,21 @@ staticfiles/
...
+static/js
+/build
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+package-lock.json
+storybook-build
+yarn.lock
```
jbothma: Nope we don't want to ignore yarn.lock - very important to keep it under version control

```suggestion
```

_includes/components/header-and-footer/Footer/index.html
```diff
@@ -0,0 +1,73 @@
+
+<footer class="Footer">
+
+    <div class="u-textAlignCenter u-marginTop20">
+      <a
+      class="Button is-secondary is-inline"
+      target="_blank"
+      href="https://docs.google.com/forms/d/e/1FAIpQLSdTgVXAPqmiT33VF_GvMtEANhxIi3V9BlVrqsHoI0Xy1Bv14w/viewform?entry.1530467677=https%3A%2F%2Fvulekamali.gov.za%2F%2F2018-19%2Ffocus%2F0.html">
```
jbothma: Hardcoding this now is going to be a problem when we do the rest of the migration. I think you should have replaced jekyll `page.url` with something like https://stackoverflow.com/questions/15436407/request-path-in-django-template

jbothma: I don't know why the old code resulted in /focus/0.html but the intention here, as per the old code, is to temlpate in the current URL

```diff
--- a/_includes/components/header-and-footer/Footer/index.html
+++ b/_includes/components/header-and-footer/Footer/index.html
@@ -5,8 +5,7 @@
       <a
       class="Button is-secondary is-inline"
       target="_blank"
-      href="https://docs.google.com/forms/d/e/1FAIpQLSdTgVXAPqmiT33VF_GvMtEANhxIi3V9BlVrqsHoI0Xy1Bv14w/viewform?entry.1530467677={{ 'https://vulekamali.gov.za/' | escape }}{{  page.url | escape }}"
-      >
+      href="https://docs.google.com/forms/d/e/1FAIpQLSdTgVXAPqmiT33VF_GvMtEANhxIi3V9BlVrqsHoI0Xy1Bv14w/viewform?entry.1530467677=https%3A%2F%2Fvulekamali.gov.za%2F%2F2018-19%2Ffocus%2F0.html">
```

budgetportal/views.py
```diff
@@ -605,3 +607,34 @@ def category_fields(category):
...
+        },
+        'site' : {
+            'data' : {
+                'videos' : read_object_from_yaml(videos_file_path),
+                'about' : read_object_from_yaml(about_date_file_path),
+                'navbar' : read_object_from_yaml(navbar_data_file_path)
+            },
+            'latest_year' : '2019-20'
+        },
+        'jekyll' : {
```
jbothma: Since we're leaving jekyll behind, the right fix for this is using the django `debug` context variable rather than defining a jekyll-specific variable in the django view.

_includes/components/header-and-footer/SubLinks/index.html
```diff
@@ -0,0 +1,39 @@
...
+      {% comment %} Determine whether children should show in sidebar {% endcomment %}
+      {% if selected_sidebar and selected_sidebar == item.id %}
+        {% assign ' is-active' as activeClass %}
+      {% else %}
+        {% assign '' as activeClass %}
+      {% endif %}
+
+      {% comment %} Determine whether link active state should be determined by JavaScript {% endcomment %}
+      {% if item.connected %}
+        {% assign ' data-connected-id='|add:item.connected as dataAttribute %}
```
jbothma: I don't understand - where do the quotes come from? Something tells me this is a hack and not a solution that's easy to maintain.

```diff
--- a/_includes/components/header-and-footer/SubLinks/index.html
+++ b/_includes/components/header-and-footer/SubLinks/index.html
@@ -21,7 +21,7 @@
 
       {% comment %} Determine whether link active state should be determined by JavaScript {% endcomment %}
       {% if item.connected %}
-        {% assign ' data-connected-id="'|add:item.connected|add:'"' as dataAttribute %}
+        {% assign ' data-connected-id='|add:item.connected as dataAttribute %}
```

_includes/components/header-and-footer/Footer/index.html
havanhuy1997: I see

budgetportal/views.py
havanhuy1997: okay, so we will use 'debug' variable

_includes/components/header-and-footer/SubLinks/index.html
havanhuy1997: I think quotes come from django template render. it auto added the quotes to value of attributes

jbothma: right - I doubt that but ok - let's see if this becomes a problem and go with this for now.

budgetportal/views.py
jbothma: yup, please go ahead with that.

_includes/components/header-and-footer/Footer/index.html
jbothma: try this

```suggestion
      href="https://docs.google.com/forms/d/e/1FAIpQLSdTgVXAPqmiT33VF_GvMtEANhxIi3V9BlVrqsHoI0Xy1Bv14w/viewform?entry.1530467677={{ request.path | iriencode }}">
```

PR #157, opened 2019-08-08
Add support for webapp package

jbothma:


budgetportal/settings.py
```diff
@@ -201,27 +202,26 @@
...
+    }
 ]
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/1.7/howto/static-files/
 STATICFILES_DIRS = [
     str(PROJ_DIR.path("static")),
-    str(ROOT_DIR.path("assets"))
+    str(ROOT_DIR.path("assets")),
+    str(ROOT_DIR.path("packages/webapp/build/static/js")),
```
jbothma: This is the only real change relevant in this file, the rest is just python black

yarn.lock
```diff
@@ -2,20 +2,13 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@7.0.0":
+"@babel/code-frame@7.0.0", "@babel/code-frame@^7.0.0":
```
jbothma: I just copied over static-budget-portal's current version over - no idea why this was different.

PR #178, opened 2019-09-22
Clean up .yaml endpoints

BarisSari:
- [x] Remove .yaml endpoints from urls.py 
- [x] Remove the .yaml endpoint tests 
- [x] Remove .yaml view code 
- [x] for views whose context function is now only used for the HTML view, merge the view function and the context function. 
- [x] Ensure there is a smoke test for each HTML view simply checking for the presence of a string that would only be present if the page loaded the data it uses correctly.

### DIFF
--- a/budgetportal/tests/test_infrastructure_projects.py
+++ b/budgetportal/tests/test_infrastructure_projects.py
@@ -1,5 +1,4 @@
 import mock
-import yaml
 from django.test import TestCase, LiveServerTestCase, Client
 
 from budgetportal.models import InfrastructureProjectPart, MAPIT_POINT_API_URL, CKAN_DATASTORE_URL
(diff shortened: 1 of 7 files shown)


budgetportal/views.py
```diff
@@ -38,50 +38,32 @@
 COMMON_DESCRIPTION_ENDING = "from National Treasury in partnership with IMALI YETHU."
 
 
-def homepage_context(request, financial_year_id):
+def homepage(request, financial_year_id="2019-20"):
     """
-    View of a financial year homepage, e.g. /2017-18
+        View of a financial year homepage, e.g. /2017-18
```
jbothma:
```suggestion
```

budgetportal/tests/test_infrastructure_projects.py
```diff
@@ -281,29 +272,25 @@ def setUp(self):
...
-        self.assertEqual(len(content['provinces']), 1)
-        self.assertIn('', content['provinces'][0])
-        self.assertEqual(content['slug'], '/infrastructure-projects/{}'.format(self.project.project_slug))
-        self.assertEqual(content['stage'], self.project.current_project_stage)
-        self.assertEqual(content['total_budget'], self.project.total_project_cost)
+        response = c.get('/infrastructure-projects/{}'.format(
+            self.project.project_slug))
+        content = response.content
+
+        self.assertTrue(content.find('fake path'))
```
jbothma: please use `assertContains` rather than `assertTrue(content.find` because assertContains is logically equivalent but gives clearer explanation what went wrong when it fails

https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.SimpleTestCase.assertContains

PR #179, opened 2019-09-23
FAQ CMS

BarisSari:
- [x] Add a model for FAQ entries 
- [x] Make the content field a Django-CKEditor RichTextField (no uploads) 
- [x] Add the current hardcoded entries as a fixture in the budgetportal app called faq.json
- [x] It should be possible to control the order that FAQ entries are shown in admin - use django-admin-sortable


budgetportal/templates/faq.html
```diff
@@ -4,44 +4,14 @@
...
-    </div>
-
-    <div class="Page-card u-marginBottom20">
-      <div class="Section">
-        <h2 class="Section-title">How does the search function work?</h2>
-        <p>The search box helps you to find budgets and datasets you may be interested in.</p>
-
-        <p>When you use the search box, we’ll try to find the most applicable budgets for your search terms. We use budget datasets and documents to help guide the results, not just department names. We do our best, but can’t always find the best budget for you. If you’re unable to find what you’re looking for please email us at feedback@vulekamali.gov.za and let us know.</p>
+        <h2 class="Section-title"> {{ faq.title }}</h2>
+        <p>{{ faq.content|safe }}</p>
```
jbothma: instead of putting this in a p, put it in a div. That way we know it'll be on a new line, but its contents could be anything. a p is only allowed certain content elements, but a div can have anything.

budgetportal/tests/test_pages.py
```diff
@@ -155,7 +155,7 @@ def test_faq_page(self):
         c = Client()
         response = c.get('/faq')
 
-        self.assertContains(response, 'When is the budget data updated?')
+        self.assertContains(response, 'Frequently Asked Questions (FAQ)')
```
jbothma: please load the faq fixture in this test class and check chat one of the FAQ titles and a snippet of its answer is in the body

budgetportal/templates/faq.html
bariss-bdt: it changed with div

https://github.com/vulekamali/datamanager/pull/179/commits/335af26cb2c506b4c9cfc516407b1d9d468b7bf7

budgetportal/tests/test_pages.py
bariss-bdt: fixture added and test updated:

https://github.com/vulekamali/datamanager/pull/179/commits/578d822f00f28fd3ca9503ec32e6469f0c36526e

PR #180, opened 2019-09-24
Video CMS

BarisSari:
- [ ] It should be possible to control the order that video entries are shown on /videos in admin - use https://github.com/alsoicode/django-admin-sortable
- [ ] It should be possible to control the order that video languages are shown using the existing inline and https://github.com/alsoicode/django-admin-sortable


jbothma:
The inline doesn't seem to be sortable for me. Does it work for you?

Screenshot_2019-09-30_10-01-33

BarisSari:
> The inline doesn't seem to be sortable for me. Does it work for you?
> 
> Screenshot_2019-09-30_10-01-33

It doesn't show sort button but you should drag&sort them. It works fine on my local project

PR #195, opened 2019-10-20
Fix Invalid Dimensions Error

BarisSari:


budgetportal/tests/test_provincial_infrastructure_projects.py
```diff
@@ -132,6 +132,19 @@ def test_assigned_correctly(self):
         self.assertEqual(principal_agent, ["BBBB"])
         self.assertEqual(other_parties, [None])
 
+    def test_empty_rows_deleted(self):
+        empty_row = [None] * self.report_sheet.data_set.width
+        num_of_current_rows = self.report_sheet.data_set.height
```
jbothma: is num_of_current_rows the number of rows with content in them?

In that case, please call it `num_nonempty_rows` otherwise the test is hard to understand.

BarisSari: yes, it is number of non-empty rows. I changed the name.

https://github.com/vulekamali/datamanager/pull/195/commits/2dfa854a62323d4e30e7b86d8ee85219e9cc383c

PR #213, opened 2019-11-21
Fix/Implement Tests for search page on snapshots and facets

BarisSari:
Not working tests:

- budgetportal.tests.test_prov_infra_projects.ProvInfraProjectWebflowIntegrationTestCase.test_number_updated_after_search

Test that is going to change to a Selenium test:
- budgetportal.tests.test_prov_infra_projects.ProvInfraProjectDetailPageTestCase


budgetportal/tests/test_bulk_upload.py
```diff
@@ -15,12 +15,6 @@
 EMAIL = "testuser@domain.com"
 PASSWORD = "12345"
 
-# Hacky make sure we don't call out to openspending.
```
jbothma: Hmmm the "hacky" came to bite us - I think we should find a better solution to this that doesn't break the tests you're busy writing, but still doesn't call out to other apps -this expands scope for you so we should discuss when and how we add that.

I think continue as is and I should just make sure to add to the scope and explain how we should handle this.

budgetportal/tests/test_irm_preprocessor.py
```diff
@@ -110,3 +110,21 @@ def test_get_row_implementors_just_one(self):
         self.assertEqual(implementors[1], "")
         self.assertEqual(implementors[2], "C")
         self.assertEqual(implementors[3], "")
+
+
+class PreprocessEmptyRowTestCase(unittest.TestCase):
+    def test_empty_rows_removed(self):
+        headers = BASE_HEADERS + [REPEATED_IMPLEMENTOR_HEADER] * 20
+        dataset = Dataset(headers=headers)
+        num_nonempty_rows = dataset.height
```
jbothma: the dataset doesn't have any nonempty rows at this point, does it?

Please add a nonempty row or two

budgetportal/tests/test_irm_preprocessor.py
```diff
@@ -110,3 +110,21 @@ def test_get_row_implementors_just_one(self):
...
+
+        num_of_empty_rows = 100
+        empty_row = [None] * len(headers)
+        for i in range(num_of_empty_rows):
+            dataset.append(empty_row)
+
+        total_rows = num_nonempty_rows + num_of_empty_rows
+        self.assertEqual(dataset.height, total_rows)
```
jbothma: please hardcode the expected values like we did before so that it's clearer when something is zero when it shouldn't be

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -31,12 +39,16 @@ def setUp(self):
...
-        super(ProvInfraProjectSeleniumTestCase, self).setUp()
-
-    def test_upload_xlsx_for_prov_infra_projects(self):
-        filename = "budgetportal/tests/test_data/test_import_prov_infra_projects.xlsx"
+        self.quarter = Quarter.objects.create(number=1)
+        self.timeout = 60
+        self.search_url = "/infrastructure-projects/provincial/"
+        super(ProvInfraProjectSeleniumIRMSnapshotTestCase, self).setUp()
+
+    def test_import_irm_snapshot(self):
```
jbothma: Tis doesn't need to be a selenium test - it's fine for it just to do IRMSnapshot.create(...) and long as it ends up querying the API

jbothma: But it's also fine as a selenium test

budgetportal/tests/test_irm_preprocessor.py
BarisSari: The expected values changed to hardcoded values.

https://github.com/vulekamali/datamanager/pull/213/commits/4b7fb0d949828d11deb02493dde613d4de3f13db

budgetportal/tests/test_irm_preprocessor.py
BarisSari: One non-empty row added.

https://github.com/vulekamali/datamanager/pull/213/commits/4b7fb0d949828d11deb02493dde613d4de3f13db

budgetportal/tests/test_prov_infra_projects.py
BarisSari: I think I'll change it to a normal test because Selenium fails even though everything seems correct.

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -49,92 +61,177 @@ def test_upload_xlsx_for_prov_infra_projects(self):
...
 
-        number_of_results = len(response.data["results"])
-        self.assertLessEqual(number_of_results, 20)
+    def tearDown(self):
+        ProvInfraProjectIndex().clear()
 
     def test_filter_by_department(self):
         department = "Department 1"
-        project = ProvInfraProject.objects.get(department=department)
+        project = ProvInfraProjectSnapshot.objects.get(department=department)
```
jbothma:
```suggestion
        project_snapshot = ProvInfraProjectSnapshot.objects.get(department=department)
```

budgetportal/tests/test_prov_infra_projects.py
BarisSari: Test changed.

https://github.com/vulekamali/datamanager/pull/213/commits/1d97757deae6ca317321058181f49a504cb5615c

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -49,92 +61,177 @@ def test_upload_xlsx_for_prov_infra_projects(self):
...
         self.statuses = ["Design", "Construction"]
         self.sources = [
             "Education Infrastructure Grant",
             "Community Library Service Grant",
         ]
+
         for i in range(30):
             if i < 15:
-                status = self.statuses[0]
+                status_ = self.statuses[0]
```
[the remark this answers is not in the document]
BarisSari: It was confused with _rest_framework.status_ . That's why I added underscore

jbothma:
i'm struggling a bit to understand the `test_facet...` tests - I think part of the issue is that the requirements weren't very clear

The idea with test descriptions like 

> - test_filter_by_funding_source
>   - update to use drf-haystack search endpoint
>   - add one for drf-haystack facet endpoint

is that _for the same database state_, a request to `search/` and `search/facets`  return consistent results - that is important because we use both endpoints to deliver results.

So the way they're written now, with different setup, they're likely to diverge.

**All the setup for these pairs of tests should be the same.**

I also worry a bit that it's not crystal clear what the setup is for the facet tests - I think it would be much easier to understand the test if it was clear what the necessary database state is. Instead of these being part of the test class where 30 projects are created, how about a test class per pair. e.g. to test that department filter works, and province facet counts work, the minimal data you need is

- project with Department A, Province P
- project with Department A, Province Q
- project with Department B, Province P

Before filtering, Province facet should have P count 2; Q count 1
After filtering Department = A, Province facet should have P count 1; Q count 1

Right now it's verify difficult to tell that the test is verifying this functionality - the connection between the data and the test isn't clear. It would be better to have slower tests with more duplication of setup and teardown code, literally going 
```
ProjectSnapshot.create(department=A, Province=P)
ProjectSnapshot.create(department=A, Province=Q)
ProjectSnapshot.create(department=B, Province=P)
...reindex()

GET ...search/
GET ...search/facet/
```
 for the above case, and repeating all for a different test checking what happens when you do a full text search, or whatever the other cases are

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -172,42 +342,150 @@ def test_filter_by_funding_source(self):
...
-        data = {"search": province}
+        data = {"q": province}
         response = self.client.get(self.url, data)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertContains(response, province)
+
+        results = response.data["results"]
+        self.assertEqual(len(results), 15)
+
+    def test_facet_search_by_province(self):
```
jbothma: `search_by_` should use  `q` to use _full text search_ rather than _filter_

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -252,14 +533,175 @@ def test_url_path(self):
...
+        url = self.project.get_absolute_url()
         response = self.client.get(url)
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertContains(response, project.name)
+        self.assertContains(response, '"status": "Completed"')
+        self.assertNotContains(response, '"status": "Construction"')
+
+    def test_latest_in_the_same_year(self):
+        latest = ProvInfraProjectSnapshot.objects.filter(project=self.project).latest()
```
jbothma:
```suggestion
        latest = self.project.project_snapshots.latest()
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -252,14 +533,175 @@ def test_url_path(self):
...
+            financial_year=self.fin_year,
+            quarter=self.quarter_1,
+            date_taken=date(year=2020, month=1, day=1),
+        )
+        self.project_snapshot_2 = ProvInfraProjectSnapshot.objects.create(
+            irm_snapshot=self.irm_snapshot_2, project=self.project,
+        )
+
+    def test_latest_in_different_years(self):
+        latest = ProvInfraProjectSnapshot.objects.filter(project=self.project).latest()
```
jbothma:
```suggestion
        latest = self.project.project_snapshots.latest()
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -252,14 +533,175 @@ def test_url_path(self):
...
+            financial_year=self.fin_year,
+            quarter=self.quarter_1,
+            date_taken=date(year=2019, month=1, day=1),
+        )
+        self.project_snapshot_1 = ProvInfraProjectSnapshot.objects.create(
+            irm_snapshot=self.irm_snapshot_1, project=self.project,
+        )
+
+        self.irm_snapshot_2 = IRMSnapshot.objects.create(
+            financial_year=self.fin_year,
```
jbothma: `financial_year` should be different - `date_taken` isn't relevant to this test so let's make it identical to demonstrate that.

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -252,14 +533,175 @@ def test_url_path(self):
...
+            irm_snapshot=self.irm_snapshot_1,
+            project=self.project,
+            status="Construction",
+        )
+
+        self.quarter_2 = Quarter.objects.create(number=2)
+        self.irm_snapshot_2 = IRMSnapshot.objects.create(
+            financial_year=self.fin_year,
+            quarter=self.quarter_2,
+            date_taken=date(year=2019, month=6, day=1),
```
jbothma: `date_taken` is irrelevant in these tests - please make them identical to demonstrate that.

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -252,14 +533,175 @@ def test_url_path(self):
...
+        response = self.client.get(self.url, data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        results = response.data["results"]
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0]["name"], "Blue School")
+        self.assertNotContains(response, "Red School")
+
+
+class ProvInfraProjectDetailPageTestCase(APITransactionTestCase):
```
jbothma:
```suggestion
class ProvInfraProjectDetailPageTestCase(TransactionTestCase):
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -252,14 +533,175 @@ def test_url_path(self):
...
+            primary_funding_source="Health Infrastructure Grant",
+            estimated_completion_date=date(year=2020, month=1, day=1),
+        )
+
+        ProvInfraProjectIndex().reindex()
+
+    def tearDown(self):
+        ProvInfraProjectIndex().clear()
+
+    def test_project_detail_page_fields(self):
```
jbothma: Is this being replaced with the webflow integration selenium tests?

budgetportal/tests/test_prov_infra_projects.py
BarisSari: Solved.

budgetportal/tests/test_prov_infra_projects.py
BarisSari: Solved.

https://github.com/vulekamali/datamanager/pull/213/commits/ce33ff33180f1d00ef558beb0e0ecbeb1f429674

budgetportal/tests/test_prov_infra_projects.py
BarisSari: I was using ```q``` but I think I forgot it in one search method. Anyway, I corrected it

budgetportal/tests/test_prov_infra_projects.py
BarisSari: yes, here it is:

https://github.com/vulekamali/datamanager/pull/213/commits/c510bce41fc8b2035a3629965510482d78688ad8

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
+        self.url = reverse("provincial-infrastructure-project-api-list")
+
+    def test_import_irm_snapshot(self):
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        # Check that 3 projects are indexed and searchable
+        results = response.data["results"]
+        num_of_results = len(results)
+        self.assertEqual(num_of_results, 3)
```
jbothma: Please create the snapshot inside the test case and assert that there are zero projects before creating it?

budgetportal/tests/test_prov_infra_projects.py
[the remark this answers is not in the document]
jbothma: Ah I see - I think a better solution is to import the status codes themselves - since the word status is quite a generic word to import by itself in the entire module. Then you don't have a conflict like this

budgetportal/tests/test_prov_infra_projects.py
jbothma: Thanks - yeah I also think it was just this one

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
         selenium = self.selenium
+        url = self.project.get_absolute_url()
+        selenium.get("%s%s" % (self.live_server_url, url))
+        title = selenium.find_element_by_css_selector(".page-heading").text
+        self.assertEqual(title, u"BLUE JUNIOR SECONDARY SCHOOL")
+
+        funding = selenium.find_element_by_xpath(
+            "//html/body/div[4]/div/div[2]/div[1]/div"
+        )
+        total_cost = funding.find_element_by_xpath(".//div[2]/div[2]").text
```
jbothma: Please use the selectors listed here for all these fields, rather than using the markup - the selectors are required to stay the same between webflow and django, so that they html can change without breaking anything.

e.g. `selenium.find_element_by_css_selector(...)`
Using XPath like this will break frequently because the HTML will change

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
-    def test_projects_per_page(self):
-        response = self.client.get(self.url, format="json")
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
+    def tearDown(self):
+        ProvInfraProjectIndex().clear()
+
+    def test_correct_numbers_showed(self):
+        selenium = self.selenium
+        selenium.get("%s%s" % (self.live_server_url, self.url))
+        num_of_projects = selenium.find_element_by_xpath(
```
jbothma: use `find_element_by_css_selector` rather than `find_element_by_xpath` - it's a lot cleaner and more like what's used in web programming

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
+            '//*[@id="num-matching-projects-field"]'
+        ).text
+        num_of_projects = int(num_of_projects)
+        self.assertEqual(num_of_projects, 11)
+
+        search_field = selenium.find_element_by_name("Infrastructure-Search")
+        search_button = selenium.find_element_by_xpath('//*[@id="Search-Button"]')
+        search_field.send_keys(province)
+        search_button.click()
+        time.sleep(self.timeout)
```
jbothma: never use time.sleep in GUI tests - rather use `wait.until`  - in this case use `text_to_be_present_in_element` as the condition https://selenium-python.readthedocs.io/waits.html

The best way to write these tests is to do
`wait.until(EC.text_to_be_present_in_element((By.CSS_SELECTOR, '#someid'), 12345))` for each check because then even if selenium is faster than our jquery, selenium will wait for each field (and it will normally just be super fast

budgetportal/tests/test_prov_infra_projects.py
jbothma: See see the comments about `wait.until` below - that will help you write cleaner, clearer and more robust tests here

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
+        est__const_end_date = const_timeline.find_element_by_xpath(
+            ".//div[4]/div[2]"
+        ).text
+
+        self.assertEqual(est_const_start_date, u"2017-02-01")
+        self.assertEqual(contracted_const_end_date, u"2021-01-01")
+        self.assertEqual(est__const_end_date, u"2020-12-31")
+
+
+class ProvInfraProjectWebflowIntegrationTestCase(BaseSeleniumTestCase):
```
jbothma:
```suggestion
class ProvInfraProjectSearchPageTestCase(BaseSeleniumTestCase):
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
             )
+        ProvInfraProjectIndex().reindex()
 
-    def test_projects_per_page(self):
-        response = self.client.get(self.url, format="json")
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
+    def tearDown(self):
+        ProvInfraProjectIndex().clear()
+
+    def test_correct_numbers_showed(self):
```
jbothma:
```suggestion
    def test_search_homepage_correct_numbers(self):
```

budgetportal/tests/test_prov_infra_projects.py
BarisSari: I moved the create to the test body.

https://github.com/vulekamali/datamanager/pull/213/commits/8e32bab500c8434735e73eea5219aaefc3969a39

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
+
         response = self.client.get(self.url, data)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
-        number_of_projects = len(response.data["results"])
-        response_data = response.data["results"][0]
-        self.assertEqual(number_of_projects, 1)
-        self.assertEqual(response_data["name"], project.name)
+        number_of_projects = response.data["count"]
+        self.assertEqual(number_of_projects, 2)
```
jbothma: You don't need to change this now, but I'm just thinking, this could be as short as

```
response = self.client.get(self.url, {"department": department})
self.assertEqual(response.status_code, status.HTTP_200_OK)
self.assertEqual(2, response.data["count"])
```

You don't have to bind everything to a variable :)

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
+
+    def test_facet_filter_by_department(self):
+        department = "Department 1"
+        province = "Eastern Cape"
+
+        response = self.client.get(self.facet_url)
+        province_facets = response.data["fields"]["province"]
+        provinces_before_filtering = 0
```
jbothma: No need to change right now, don't think an error here would be too confusing, but as a best practise I tend to initialise values to None rather than something that might actually come back from a valid function call, if I have to initialise functions at all

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
+        for value in province_facets:
+            if province == value["text"]:
+                provinces_before_filtering = value["count"]
```
jbothma:
```suggestion
                province_projects_before_filtering = value["count"]
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,959 @@
...
+        number_of_projects = response.data["count"]
+        self.assertEqual(number_of_projects, 2)
+
+    def test_facet_filter_by_province(self):
+        province = "Eastern Cape"
+        department = "Department 1"
+
+        response = self.client.get(self.facet_url)
+        department_facets = response.data["fields"]["department"]
+        departments_before_filtering = 0
```
jbothma:
```suggestion
        department_projects_before_filtering = 0
```

BarisSari: Names changed.

https://github.com/vulekamali/datamanager/pull/213/commits/25ace2b66563dbb756156f9ddca0c1d3dfb9b381

budgetportal/tests/test_prov_infra_projects.py
BarisSari: Names changed.

https://github.com/vulekamali/datamanager/pull/213/commits/25ace2b66563dbb756156f9ddca0c1d3dfb9b381

budgetportal/tests/test_prov_infra_projects.py
BarisSari: wait.until added.

https://github.com/vulekamali/datamanager/pull/213/commits/0809f4566bfa62d3e81e546bbd15d4096f5a42d5

budgetportal/tests/test_prov_infra_projects.py
BarisSari: all find_element_by_xpath methods changed with find_element_by_css_selector

budgetportal/tests/test_prov_infra_projects.py
BarisSari: all find_element_by_xpath methods changed with find_element_by_css_selector

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,1041 @@
...
+        self.assertEqual(num_of_projects, 11)
+
+        search_field = selenium.find_element_by_css_selector(
+            "#Infrastructure-Search-Input"
+        )
+        search_button = selenium.find_element_by_css_selector("#Search-Button")
+        search_field.send_keys(province)
+        search_button.click()
+        wait = WebDriverWait(selenium, self.timeout)
+        wait.until(
```
jbothma: it doesn't help to do wait.until for something that is already on the page.

The point of using wait.until is so that selenium waits until something that should be updated by javascript has been updated.

In this case, you are waiting until a heading which is built into the html is on the page after you clicked a button - that button won't reload the page, so it's immediately successful - so this doesn't end up waiting for the asynchronous javascrip stuff to complete.

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,1041 @@
...
+            )
+        )
 
-        number_of_results = len(response.data["results"])
-        self.assertLessEqual(number_of_results, 20)
+        filtered_num_of_projects = selenium.find_element_by_css_selector(
+            "#num-matching-projects-field"
+        ).text
+        filtered_num_of_projects = int(filtered_num_of_projects)
+        self.assertEqual(filtered_num_of_projects, 5)
```
jbothma: this is where you want to do a wait.until - in one expression, you can say wait until #num-matching-projects-field contains "5" with the example I gave.

That takes care of waiting for javascript, finding the element, comparing the value, and raising an exception if it doesn't happen within the timeout. It also polls, so it should be pretty quick, rather than sleeping 5 seconds and then checking.

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,1041 @@
...
+
+    def test_number_updated_after_search(self):
+        province = "Eastern Cape"
+        selenium = self.selenium
+        selenium.get("%s%s" % (self.live_server_url, self.url))
+        num_of_projects = selenium.find_element_by_css_selector(
+            "#num-matching-projects-field"
+        ).text
+        num_of_projects = int(num_of_projects)
+        self.assertEqual(num_of_projects, 11)
```
jbothma: this should be done with wait.until so that selenium waits for the javascript that inserts this value executes and updates the page.

jbothma: It might be working now, but this is the kind of thing that fails sometimes, with no explanation why - because it's asynchronous and sometimes python is faster and sometimes javascript is faster. So the best practise is to use this wait-until pattern with an appropriate condition. People who don't do this end up thinking that selenium is unreliable - selenium is great!

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,247 +1,1041 @@
...
+    def tearDown(self):
+        ProvInfraProjectIndex().clear()
+        self.file.close()
 
+    def test_project_detail_page_fields(self):
         selenium = self.selenium
+        url = self.project.get_absolute_url()
+        selenium.get("%s%s" % (self.live_server_url, url))
+        title = selenium.find_element_by_css_selector(".page-heading").text
+        self.assertEqual(title, u"BLUE JUNIOR SECONDARY SCHOOL")
```
jbothma: please use wait.until for this, at least for the first one

PR #215, opened 2019-11-26
Prepare project time series chart data

BarisSari:


budgetportal/prov_infra_project/charts.py
```diff
@@ -0,0 +1,135 @@
...
+            chart_data = update_previous_chart_values(
+                chart_data, snapshot, quarter_number, fin_year
+            )
+    chart_data["snapshots"] = order_chart_data(chart_data["snapshots"])
+
+    latest_snapshot = find_latest_snapshot_with_dates(project)
+    if latest_snapshot:
+        chart_data["events"] = extract_events(latest_snapshot)
+
+    return json.dumps(chart_data, use_decimal=True)
```
jbothma: this should just return the chart_data object - the caller will encode as json. But the output of this should be nested dictionaries and lists that are compativle with json.dumps. So it looks like we should have a test that verifies that the output of this function can be passed to `json.dumps`

Why are you using simplejson? The task was to use `json.dumps` - by that I meant the json package built into python.

I'm using a customised encoder to format dates in `budgetportal.webflow.views`

jbothma: Perhaps it's a good time to move the encoder to a module like `budgetportal.json` then it can be used in this test and in the webflow views.

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,817 @@
...
+        )
+        self.project_snapshot_2 = ProvInfraProjectSnapshot.objects.create(
+            irm_snapshot=irm_snapshot_2, project=self.project, status="Tender"
+        )
+
+    def tearDown(self):
+        self.file.close()
+
+    def test_status_is_null(self):
+        """Test that status for Q1 (which created by Q2) is Null"""
```
jbothma:
```suggestion
        """Test that status for Q1 (which created by Q2 snapshot) is Null"""
```

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,817 @@
...
+        )
+        self.project_snapshot_2 = ProvInfraProjectSnapshot.objects.create(
+            irm_snapshot=irm_snapshot, project=self.project, total_project_cost=100
+        )
+
+    def tearDown(self):
+        self.file.close()
+
+    def test_total_project_cost_is_null(self):
+        """Test that total project cost for Q1 (which created by Q2) is Null"""
```
jbothma:
```suggestion
        """Test that total project cost for Q1 (which created by Q2 snapshot) is Null"""
```

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,817 @@
...
+            actual_expenditure_q1=10,
+            expenditure_from_previous_years_total=200,
+        )
+
+    def tearDown(self):
+        self.file.close()
+        self.file_2.close()
+
+    def test_q1_updated_after_q2_snapshot_inserted(self):
+        """Test that Q1 is updated correctly when Q2 inserted"""
```
jbothma:
```suggestion
        """Test that Q1 values are updated correctly when Q2 snapshot is added"""
```

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,817 @@
...
+        # Check Q1 values
+        self.assertEqual(snapshots_data[0]["total_spent_in_quarter"], 12)
+        self.assertEqual(snapshots_data[0]["total_spent_to_date"], 212)
+
+        # Check Q2 values
+        self.assertEqual(snapshots_data[1]["total_spent_in_quarter"], 21)
+        self.assertEqual(snapshots_data[1]["total_spent_to_date"], 233)
+
+
+class Q1Q2Q3UpdateTestCase(TransactionTestCase):
```
jbothma: please delete - I think the previous two are sufficient to prove the point

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,817 @@
...
+            expenditure_from_previous_years_total=100,
+        )
+
+    def tearDown(self):
+        self.file.close()
+        self.file_2.close()
+
+    def test_correct_value_used_for_previous_total(self):
+        """
+        Test that Q2's expenditure_from_previous_years_total changes total_spent of Q1.
```
jbothma:
```suggestion
        Test that Q2 snapshot's expenditure_from_previous_years_total updates total_spent of Q1 chart item.
```

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,846 @@
...
+        snapshots_data = time_series_data(self.project.project_snapshots.all())
+        snapshots_data = snapshots_data[u"snapshots"]
+        self.assertEqual(len(snapshots_data), 2)
+
+        # Check total_spent_to_date values for Q1 and Q2
+        self.assertEqual(snapshots_data[0]["total_spent_to_date"], None)
+        self.assertEqual(snapshots_data[1]["total_spent_to_date"], None)
+
+
+class NullQuarterlySpendSecondTestCase(TransactionTestCase):
```
jbothma:
```suggestion
class NullQ2SubsequentNullSpendTestCase(TransactionTestCase):
```

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,846 @@
...
+            quarter=q1,
+            date_taken=date(2030, 6, 30),
+            file=File(self.file),
+        )
+        ProvInfraProjectSnapshot.objects.create(
+            irm_snapshot=irm_snapshot_1,
+            project=self.project,
+            actual_expenditure_q1=10,
+            expenditure_from_previous_years_total=200,
+        )
```
jbothma: I think you only need the second snapshot for this test because we're testing subsequent nulls, not updating.

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,846 @@
...
+            quarter=q1,
+            date_taken=date(2030, 6, 30),
+            file=File(self.file),
+        )
+        ProvInfraProjectSnapshot.objects.create(
+            irm_snapshot=irm_snapshot_1,
+            project=self.project,
+            actual_expenditure_q1=None,
+            expenditure_from_previous_years_total=200,
+        )
```
jbothma: I think you only need the second snapshot for this test - we're testing subsequent nulls, not updates.

In fact, I think we only need one test case, not two: a Q3 or Q4 snapshot where Q1 and Q3 actuals are set and Q2 actual is null - it can show that Q1 spend to date is calculated and the rest are null.

budgetportal/tests/test_prov_infra_chart.py
BarisSari: Deleted.

https://github.com/vulekamali/datamanager/pull/215/commits/e48f74125c5e7c0745191f6801f5c0e1d70f9171

budgetportal/prov_infra_project/charts.py
```diff
@@ -0,0 +1,131 @@
...
+            return snapshot
+
+    return None
+
+
+def extract_events(project_snapshot):
+    events = []
+    start_date = project_snapshot.estimated_construction_start_date
+    events.append(
+        {"date": str(start_date), "label": "Estimated construction start date"}
```
jbothma: > - All the date fields in the snapshots for which the latest snapshot of all snapshots is not null should be output as elements of the events array using the column name in the excel files passed to IRMSnapshot as their label.
>   - hardcode the field names and labels - don't try and introspect and generate these based on the type of field - I'm just using the fact that they're all currently date fields to simplify this description.

```
    start_date = models.DateField(blank=True, null=True)
    estimated_construction_start_date = models.DateField(blank=True, null=True)
    estimated_completion_date = models.DateField(blank=True, null=True)
    contracted_construction_end_date = models.DateField(blank=True, null=True)
    estimated_construction_end_date = models.DateField(blank=True, null=True)
```

jbothma: Please use the label `Project Start Date` for  `start_date`

budgetportal/tests/test_prov_infra_chart.py
```diff
@@ -0,0 +1,735 @@
...
+        self.project_snapshot_2 = ProvInfraProjectSnapshot.objects.create(
+            irm_snapshot=irm_snapshot_2, project=self.project
+        )
+
+    def tearDown(self):
+        self.file.close()
+
+    def test_two_snapshots_emitted(self):
+        """Test that Q2 created 2 items"""
```
jbothma:
```suggestion
        """Test that if the first snapshot is Q2, an item is also created for Q1"""
```

budgetportal/tests/test_prov_infra_chart.py
BarisSari: https://github.com/vulekamali/datamanager/pull/215/files/8003845d3d85a9e61ca08ad7c6fc0f02fe05b219#diff-93488925aca732879663affbdf6b3505

budgetportal/tests/test_prov_infra_chart.py
BarisSari: https://github.com/vulekamali/datamanager/pull/215/files/8003845d3d85a9e61ca08ad7c6fc0f02fe05b219#diff-93488925aca732879663affbdf6b3505

budgetportal/prov_infra_project/charts.py
```diff
@@ -0,0 +1,131 @@
+
+def order_chart_data(snapshot_list):
+    return sorted(snapshot_list, key=lambda x: x["date"])
+
+
+def time_series_data(project_snapshots):
```
jbothma: Let's just document that this expects a collection of ProvInfraProjectSnapshots of the same project, in IRMSnapshot order.

budgetportal/prov_infra_project/charts.py
BarisSari: All date fields added to extract_events

https://github.com/vulekamali/datamanager/pull/215/commits/8c598af2001e62b60960b89ddf88e9461cc19e5a

budgetportal/prov_infra_project/charts.py
BarisSari: Encoder moved.

https://github.com/vulekamali/datamanager/pull/215/commits/f4a9dc067dbf8387d215c359592855f043b29d4d

budgetportal/prov_infra_project/charts.py
BarisSari: Docstring added

https://github.com/vulekamali/datamanager/pull/215/commits/821d94b299019a7985f9ac13a832e99b691b10f0

PR #223, opened 2020-01-31
Humanise amounts

jbothma:


jbothma:
I think this is ready to deploy now https://trello.com/c/sqNusniT/352-get-prov-infra-projects-live

Will do it on the demo site and then production on monday

PR #260, opened 2020-02-29
Add provincial project search and detail csv download urls

TomaszKolek:


TomaszKolek:
@jbothma its totally WIP PR still, some things are committed but shouldn't (like template change)
If you have a while - I'd be happy to read your opinion about APIs, model, and approach at all.

budgetportal/fixtures/provincial-infrastructure-projects.json
```diff
@@ -1,6 +1,6 @@
 [
 {
-  "model": "budgetportal.provinfraproject",
+  "model": "budgetportal.provinfraprojectsnapshot",
```
TomaszKolek: Let me please know if it does make sense, I think this fixture is a bit outdated but maybe I'm wrong.

budgetportal/management/commands/load_departments.py
```diff
@@ -79,7 +79,10 @@ def create_dept():
                     )
 
                 try:
-                    if update:
+                    exists = Department.objects.filter(
```
TomaszKolek: the script is not working when update is set to False but the department exists. I think we could get rid of the update flag and just check if dep exists.

budgetportal/webflow/templates/webflow/detail_provincial-infrastructure-projects.html
```diff
@@ -177,7 +177,7 @@ <h1 class="page-heading name-field"></h1>
     <div class="container">
       <div class="row project-info__header">
         <div class="col">
-          <h2 class="sub-section__title">Project maps</h2>
+          <h2 class="sub-section__title">Project maps</h2> <p>{{ download_url }}</p>
```
TomaszKolek: testing purposes

budgetportal/webflow/urls.py
```diff
@@ -21,5 +22,15 @@
...
     ),
+    url(
+        r"^infrastructure-projects/provincial/(?P<id>\d+)-(?P<slug>[\w-]+)/csv-download$",
+        views.ProvInfaProjectCSVDownload.as_view(),
+        name="provincial-infra-project-detail-csv-download",
+    ),
     url(r"^api/v1/", include(router.urls)),
+    url(
+        r"^api/v1/infrastructure-projects/provincial/search/csv-download/<uuid:uuid>",
+        views.ProvInfraProjectSearchViewCSVDownload.as_view(),
```
TomaszKolek: decided to make it here not as a view set as its just one get action - let me know if it does make sense

budgetportal/webflow/views.py
```diff
@@ -60,12 +64,41 @@ def provincial_infrastructure_project_detail(request, id, slug):
         % (snapshot.name, snapshot.province),
         "page_description": "Provincial infrastructure project by the %s %s department."
         % (snapshot.province, snapshot.department),
+        "download_url": project.csv_download_url
```
TomaszKolek: the name is probably not descriptive enough

budgetportal/webflow/urls.py
```diff
@@ -21,5 +22,15 @@
...
     ),
+    url(
+        r"^infrastructure-projects/provincial/(?P<id>\d+)-(?P<slug>[\w-]+)/csv-download$",
+        views.ProvInfaProjectCSVDownload.as_view(),
+        name="provincial-infra-project-detail-csv-download",
+    ),
     url(r"^api/v1/", include(router.urls)),
+    path(
+        "infrastructure-projects/provincial/search/get-csv",
+        views.ProvInfraProjectSearchView.as_view({"get": "get_csv"}),
```
TomaszKolek: here is the view set thing. We still use the same class that is cool, we do not need to worry about having the same filters etc

budgetportal/webflow/views.py
```diff
@@ -104,6 +140,55 @@ def __init__(self, *args, **kwargs):
                 self.fields.pop(field_name)
 
 
+class ProvInfraProjectCSVSerializer(HaystackSerializer):
+    class Meta:
+        index_classes = [ProvInfraProjectIndex]
+        fields = [
```
TomaszKolek: Unfortunately Haystack does not support "__all__" option

budgetportal/fixtures/provincial-infrastructure-projects.json
jbothma: is any test using it? how does this even work?

budgetportal/webflow/urls.py
```diff
@@ -21,5 +22,15 @@
         views.provincial_infrastructure_project_detail,
         name="provincial-infra-project-detail",
     ),
+    url(
+        r"^infrastructure-projects/provincial/(?P<id>\d+)-(?P<slug>[\w-]+)/csv-download$",
```
jbothma: now that it's django 2 thanks to you, you can do something like `<int:id>-<slug:slug>`

budgetportal/webflow/urls.py
```diff
@@ -21,5 +22,15 @@
...
+        views.ProvInfaProjectCSVDownload.as_view(),
+        name="provincial-infra-project-detail-csv-download",
+    ),
     url(r"^api/v1/", include(router.urls)),
+    path(
+        "infrastructure-projects/provincial/search/get-csv",
```
jbothma: Let's make it

```suggestion
        "infrastructure-projects/provincial/search/csv",
```

budgetportal/webflow/views.py
```diff
@@ -164,3 +250,26 @@ class ProvInfraProjectSearchView(FacetMixin, HaystackViewSet):
...
+        return super().list(request, *args, **kwargs)
+
+    def _get_csv_query_params(self, original_query_params):
+        csv_download_params = original_query_params.copy()
+        csv_download_params.pop("fields", None)
+        csv_download_params.pop("ordering", None)
+        csv_download_params.pop("limit", None)
+        params = ""
+        for param_name, param_value in csv_download_params.items():
+            params += "{}={}".format(param_name, param_value)
```
jbothma: please use urllib.parse.urlencode

budgetportal/fixtures/provincial-infrastructure-projects.json
TomaszKolek: I do not think so, I just used it as part of my initial loading data tries (before you sent those files to me)

budgetportal/tests/test_pages.py
```diff
@@ -228,7 +228,7 @@ def test_dataset_category_page(self):
 
         self.assertContains(response, "<title>test-slug - vulekamali</title>")
         self.assertContains(response, "basic-test-description")
-        self.assertContains(response, "Learn more about test-slug")
+        # self.assertContains(response, "Learn more about test-slug")
```
TomaszKolek: I checked this and this test failure is not probably related to my code changes - please take a look. based on dicts we have hardcoded there is no way to find test-slug there. `guide` variable will be None in such case.

TomaszKolek: Actually didn't try rebase - will check it out at the end

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*
 import os
+import io
+import csv
 from datetime import date
-
+from slugify import slugify
```
TomaszKolek: this import can be deleted probably.

docker-compose.yml
```diff
@@ -93,7 +93,7 @@ services:
     command: minio --compat server data
 
   solr:
-    build: "https://github.com/vulekamali/solr-dokku.git"
+    build: "https://github.com/TomaszKolek/solr-dokku.git"
```
TomaszKolek: This needs to be deleted before merged, also PR that is created in main solr-dokku repo needs to be merged.

budgetportal/fixtures/provincial-infrastructure-projects.json
TomaszKolek: I checked this - its not being used.

budgetportal/tests/test_pages.py
TomaszKolek: Confirmed, that test is buggy also on muster
https://github.com/vulekamali/datamanager/pull/262

docker-compose.yml
TomaszKolek: https://github.com/vulekamali/solr-dokku/pull/7

budgetportal/fixtures/provincial-infrastructure-projects.json
jbothma: Thanks - I added it to the dev setup docs.

budgetportal/models.py
```diff
@@ -9,6 +9,7 @@
 from itertools import groupby
 from pprint import pformat
 from urllib.parse import urljoin, quote
+from uuid import uuid4
```
jbothma: we don't need this any more, do we?

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -49,6 +55,10 @@ def test_import_irm_snapshot(self):
         results = response.data["results"]
         num_of_results = len(results)
         self.assertEqual(num_of_results, 0)
+        self.assertEqual(
+            response.data["csv_download_url"],
+            reverse("provincial-infrastructure-project-api-csv"),
```
jbothma: I would prefer this to be hardcoded so that it's super explicit - otherwise we're just testing that this function is used in the view, and that if the url is changed without meaning to we won't notice (but links to the site might break).

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -63,6 +73,10 @@ def test_import_irm_snapshot(self):
         results = response.data["results"]
         num_of_results = len(results)
         self.assertEqual(num_of_results, 3)
+        self.assertEqual(
+            response.data["csv_download_url"],
+            reverse("provincial-infrastructure-project-api-csv"),
```
jbothma: same here

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1212,3 +1226,204 @@ def test_correct_project_returned(self):
         self.assertEqual(len(results), 1)
         self.assertEqual(results[0]["name"], "Blue School")
         self.assertNotContains(response, "Red School")
+        self.assertEqual(
+            response.data["csv_download_url"],
+            reverse("provincial-infrastructure-project-api-csv")
```
jbothma: and here

budgetportal/webflow/urls.py
```diff
@@ -11,15 +12,25 @@
 )
 
 urlpatterns = [
-    url(
-        r"^infrastructure-projects/provincial/$",
+    path(
+        r"infrastructure-projects/provincial/",
```
jbothma: out of curiosity, why relax this?

budgetportal/models.py
TomaszKolek: True, to remove.

budgetportal/tests/test_prov_infra_projects.py
TomaszKolek: Makes sense.

budgetportal/webflow/urls.py
TomaszKolek: do you mean why I removed ^ and $? Its now by default in path.

budgetportal/webflow/views.py
```diff
@@ -60,12 +67,73 @@ def provincial_infrastructure_project_detail(request, id, slug):
...
+        headers = self._get_headers(ProvInfraProjectCSVSerializer.Meta.fields)
+        writer = csv.DictWriter(Echo(), fieldnames=headers)
+        yield writer.writerow({h: h for h in headers})
+
+        for row in response_results:
+            prepared_row = self._prepare_row(row)
+            yield writer.writerow(prepared_row)
+
+    def _get_headers(self, headers):
+        return [self.labels.get(header, header) for header in headers]
```
jbothma: if you have time, fixing this field name in the model (mapping at import-time) and index would be preferable

budgetportal/webflow/views.py
```diff
@@ -60,12 +67,73 @@ def provincial_infrastructure_project_detail(request, id, slug):
...
+
+    def _prepare_row(self, row):
```
jbothma: same as https://github.com/vulekamali/datamanager/pull/260/files#r387645901

budgetportal/tests/test_pages.py
jbothma: thanks - fixed master and merged.

budgetportal/webflow/urls.py
jbothma: ah cool - thanks - and thanks for updating the others with slug:slug for consistency

budgetportal/webflow/views.py
jbothma: I meant for the `"{}={}"` you pass in a dictionary and it does the key=value

jbothma: instead of doing `"{}={}"``and `&` yourself, it takes care of all of that

budgetportal/webflow/views.py
```diff
@@ -164,3 +284,45 @@ class ProvInfraProjectSearchView(FacetMixin, HaystackViewSet):
...
+            "department",
+            "status",
+            "primary_founding_source",
+            "q",
+        )
+        extension = "csv"
+        filename = "provincial-infrastructure-projects"
+        for key in keys_to_check:
+            if query_params.get(key):
+                filename += "-{}-{}".format(key, slugify(query_params[key]))
```
jbothma: slugify key as well please (in case someone crafts a weird url)

budgetportal/webflow/views.py
```diff
@@ -164,3 +284,45 @@ class ProvInfraProjectSearchView(FacetMixin, HaystackViewSet):
...
+            response.data["results"], filename=self._get_filename(request.query_params)
+        )
+
+    def _get_csv_query_params(self, original_query_params):
+        csv_download_params = original_query_params.copy()
+        csv_download_params.pop("fields", None)
+        csv_download_params.pop("limit", None)
+        params = ""
+        for param_name, param_value in csv_download_params.items():
+            params += "{}={}&".format(param_name, urllib.parse.quote(param_value))
```
jbothma: see https://github.com/vulekamali/datamanager/pull/260#pullrequestreview-367245834

budgetportal/webflow/views.py
```diff
@@ -164,3 +284,45 @@ class ProvInfraProjectSearchView(FacetMixin, HaystackViewSet):
...
+        return params[:-1]
```
jbothma: why `[:-1]`?

budgetportal/models.py
TomaszKolek: Not it cannot be  `irm_snapshot_file_path` uses it

budgetportal/webflow/views.py
TomaszKolek: we use only particular keys so do not need to slugify it.
Let me know if we want to change that and use all keys from url.

budgetportal/webflow/views.py
TomaszKolek: we do not need last & actually.

budgetportal/webflow/views.py
TomaszKolek: right, it fits here well.

jbothma:
we get the `offset` query parameter in the csv url for all but the first search coordinates API request
Screenshot_2020-03-04_17-09-30

jbothma:
We only get 20 results in the search download even when there are more on the page

budgetportal/webflow/views.py
```diff
@@ -60,12 +67,50 @@ def provincial_infrastructure_project_detail(request, id, slug):
         % (snapshot.name, snapshot.province),
         "page_description": "Provincial infrastructure project by the %s %s department."
         % (snapshot.province, snapshot.department),
+        "csv_download_url": project.csv_download_url,
```
jbothma: this should be in the page_data variable so that it shows up in the embedded json for javascript to use
Screenshot_2020-03-04_17-06-09

TomaszKolek: Makes sense to me

TomaszKolek:
Going to skip the offset also dynamically change default page size - should work that way

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -13,19 +13,17 @@
 )
 from budgetportal.search_indexes import ProvInfraProjectIndex
 from budgetportal.tests.helpers import BaseSeleniumTestCase
+from django.conf import settings
 from django.core.files import File
 from django.urls import reverse
 from rest_framework import status
-from rest_framework.test import APITransactionTestCase
+from rest_framework.test import APITransactionTestCase, override_settings
```
TomaszKolek: override_settings not needed

budgetportal/fixtures/infrastructure-projects-2019-20.csv
```diff
@@ -1,4 +1,4 @@
-Sphere,department,sector,project_name,project_description,gps_code,nature_of_investment,infrastructure_type,current_project_stage,sip_category,total_project_cost,br_featured,featured,financial_year,budget_phase,amount,project_slug
+Sphere,department,sector,project_name,project_description,gps_code,nature_of_investment,infrastructure_type,current_project_stage,sip_category,estimated_total_project_cost,br_featured,featured,financial_year,budget_phase,amount,project_slug
```
jbothma: this is for national projects, not provincial - please revert

budgetportal/fixtures/test-infrastructure-pages-detail.json
```diff
@@ -18,7 +18,7 @@
       "project_slug": "basic-test-slug",
       "amount": 1368285000,
       "financial_year": 2015,
-      "total_project_cost": 9045389000,
+      "estimated_total_project_cost": 9045389000,
```
jbothma: this is for national infrastructure projects - not related - please revert

budgetportal/fixtures/test-infrastructure-pages-overview.json
```diff
@@ -18,7 +18,7 @@
       "project_slug": "health-standard-fake-project",
       "amount": 1368285000,
       "financial_year": 2015,
-      "total_project_cost": 9045389000,
```
jbothma: this is national projects, not provincial - please revert

budgetportal/management/commands/load_departments.py
jbothma: I think that's the point? that it only updates when you explicitly say to update.

budgetportal/models.py
jbothma: why is it showing as a new line in this diff then? :)

jbothma: irm_snapshot_file_path uses the `uuid` import higher up - unless that's changed, in which case you can delete the bare uuid import

budgetportal/fixtures/infrastructure-projects-2019-20.csv
TomaszKolek: omg ... sorrry, sure

TomaszKolek:
Sorry @jbothma it was too late for such changes yesterday ;-)

budgetportal/webflow/views.py
jbothma: oh I see - ok cool

budgetportal/migrations/0044_auto_20200305_0938.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("budgetportal", "0043_auto_20200304_1335"),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name="infrastructureprojectpart",
```
jbothma: we don't want to be renaming this - it's also national - please remove and I'll roll back this one on my side

TomaszKolek: This migration rolls it back. Going to squash those migrations once tests pass.

TomaszKolek: by squash I mean remove those 2 renamings

budgetportal/migrations/0043_auto_20200304_1335.py
```diff
@@ -0,0 +1,40 @@
...
+            model_name="provinfraprojectsnapshot",
+            old_name="adjustment_appropriation_professional_fees",
+            new_name="adjusted_appropriation_professional_fees",
+        ),
+        migrations.RenameField(
+            model_name="provinfraprojectsnapshot",
+            old_name="adjustment_appropriation_construction_costs",
+            new_name="adjusted_appropriation_construction_costs",
+        ),
+        migrations.RenameField(
```
TomaszKolek: this needs to be removed

budgetportal/views.py
```diff
@@ -557,7 +557,7 @@ def infrastructure_project_detail_data(project_slug):
         "stage": project.current_project_stage,
         "description": project.project_description,
         "provinces": project.provinces.split(","),
-        "total_budget": project.total_project_cost,
+        "total_budget": project.estimated_total_project_cost,
```
jbothma: this is national - revert please

budgetportal/views.py
```diff
@@ -492,7 +492,7 @@ def infrastructure_projects_overview(request):
                 "stage": project.current_project_stage,
                 "description": project.project_description,
                 "provinces": project.provinces.split(","),
-                "total_budget": project.total_project_cost,
+                "total_budget": project.estimated_total_project_cost,
```
jbothma: this is national - revert please

budgetportal/management/commands/load_departments.py
TomaszKolek: but if you do not say that and your Department exists it breaks, maybe we should just skip the iteration when Dep exists and there is no update flag?

jbothma: can you just revert this for this PR? sorry I just don't want to think about this right now - I'm not sure why you need to be running this twice and there's too much else going on for me to think about this properly

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1212,3 +1224,248 @@ def test_correct_project_returned(self):
...
+        )
+        ProvInfraProjectIndex().reindex()
+
+    def tearDown(self):
+        ProvInfraProjectIndex().clear()
+        self.file1.close()
+        self.file2.close()
+
+    def test_csv_download_empty_file(self):
+        data = {"q": "data that won't be found"}
```
jbothma:
```suggestion
    def test_csv_download_empty_file(self):
        """
        Verifies 1) correct filename, 2) header but zero rows of data present
        """
        data = {"q": "data that won't be found"}
```

jbothma: Please also check header and number of rows

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1212,3 +1224,248 @@ def test_correct_project_returned(self):
...
+        )
+
+        content = b"".join(response.streaming_content)
+        csv_reader = csv.DictReader(io.StringIO(content.decode("utf-8")))
+        self.assertListEqual(
+            csv_reader.fieldnames, ProvInfraProjectCSVSerializer.Meta.fields
+        )
+
+    def test_csv_download_with_all_projects(self):
+        response = self.client.get(self.url)
```
jbothma:
```suggestion
    def test_csv_download_with_all_projects(self):
        """
        Verifies that 1) correct filename, 2) correct header, number of rows for all projects
        """
        response = self.client.get(self.url)
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1212,3 +1224,248 @@ def test_correct_project_returned(self):
         self.assertEqual(len(results), 1)
         self.assertEqual(results[0]["name"], "Blue School")
         self.assertNotContains(response, "Red School")
+        self.assertEqual(
+            response.data["csv_download_url"],
+            "/infrastructure-projects/provincial/search/csv?q=Eastern+Cape+School",
+        )
+
+
+class ProvInfraProjectIRMSnapshotCSVDownloadMixin:
```
jbothma:
```suggestion
class TestProvInfraProjectSearchCSVMixin:
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1212,3 +1224,248 @@ def test_correct_project_returned(self):
...
+    def _test_csv_content_correctness(self, csv_reader, items_to_compare):
+        for index, row in enumerate(csv_reader):
```
jbothma:
```suggestion
    def _test_csv_content_correctness(self, csv_reader, items_to_compare, expected_num_rows):
        self.assertEqual(expected_num_rows, csv_reader)
        for index, row in enumerate(csv_reader):
```

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1212,3 +1224,248 @@ def test_correct_project_returned(self):
...
+            self.assertListEqual(
+                list(row.keys()), ProvInfraProjectCSVSerializer.Meta.fields
+            )
```
jbothma:
```suggestion
            # Verify all the serializer fields are present in the CSV
            self.assertListEqual(
                list(row.keys()), ProvInfraProjectCSVSerializer.Meta.fields
            )
```

jbothma: Please add comments for what the rest of the assertions verify since the abstraction makes it hard to know exactly what the test is checking.

I find that number 1 priority for tests is to easily see what is being tested (and what isn't) without executing code in your head

budgetportal/management/commands/load_departments.py
TomaszKolek: 🆗

budgetportal/migrations/0044_auto_20200305_0938.py
jbothma: hah - sorry, I was also tired :)

budgetportal/tests/test_prov_infra_projects.py
```diff
@@ -1212,3 +1224,301 @@ def test_correct_project_returned(self):
...
+
+            # Verify correctness of the string representation of irm_snapshot between CSV and model
+            self.assertEqual(
+                str(items_to_compare[index].irm_snapshot), row["irm_snapshot"]
+            )
+
+            # Verify correctness of province field of project between CSV and model
+            self.assertEqual(items_to_compare[index].province, row["province"])
+
+            # Verify correctness of province field of project between CSV and model
```
jbothma:
```suggestion
            # Verify correctness of completion date field of project between CSV and model
```

docker-compose.yml
jbothma: merged
```suggestion
    build: "https://github.com/vulekamali/solr-dokku.git"
```

PR #270, opened 2020-03-10
Wagtail

TomaszKolek:


budgetportal/migrations/0043_auto_20200310_1720.py
```diff
@@ -0,0 +1,97 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailimages', '0001_squashed_0021'),
+        ('wagtailcore', '0045_assign_unlock_grouppagepermission'),
+        ('budgetportal', '0042_auto_20200131_1500'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
TomaszKolek: List of unapplied migrations - I believe someone just made it directly on DB (sql) or changed things like help text.
Easy to remove in case its not good, but its a bit annoying to remove it everytime

budgetportal/management/commands/flush.py
```diff
@@ -0,0 +1,10 @@
+from django.core.management.commands.flush import Command as BaseFlushCommand
+
+
+class Command(BaseFlushCommand):
+    def add_arguments(self, parser):
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--allow-cascade', action='store_true', dest='allow_cascade', default=True,
+                help='Adds "CASCADE" option to TRUNCATE command if supported by db backend. Default=False.',
+        )
```
TomaszKolek: This is definitely something we need to think about.

budgetportal/tests/helpers.py
```diff
@@ -45,3 +46,21 @@ def wait_until_text_in(self, selector, text):
         self.wait.until(
             EC.text_to_be_present_in_element((By.CSS_SELECTOR, selector), text)
         )
+
+    def _fixture_teardown(self):
```
TomaszKolek: Ugly as f*** but probably the only one simple workaround for that issue.

budgetportal/tests/helpers.py
```diff
@@ -45,3 +46,41 @@ def wait_until_text_in(self, selector, text):
...
+                            hasattr(connections[db_name], '_test_serialized_contents')
+                    )
+            )
+            call_command('flush', verbosity=0, interactive=False,
+                         database=db_name, reset_sequences=False,
+                         allow_cascade=True,
+                         inhibit_post_migrate=inhibit_post_migrate)
+
+
+class CustomLiveServerTestCase(LiveServerTestCase):
```
TomaszKolek: Its probably not needed as tests that used `LiveServerTestCase` could be simple APITestCase IMHO.
But lets leave it as the last part. It was really time-consuming to discover those things.

PR #271, opened 2020-03-11
Wagtail (try 2)

TomaszKolek:


budgetportal/models/pages.py
```diff
@@ -0,0 +1,145 @@
...
+    max_count = 1
+
+
+class LearningIndexPage(RoutablePage):
+    max_count = 1
+
+
+class PostIndexPage(RoutablePage):
+    max_count = 1
+    parent_page_types = []
```
jbothma: I don't think we want this any more - this makes it impossible to create the page - perhaps it's ok if you create it using a fixture, but I think max_count will make sure only one can be created in any case (and like this, it can't be re-created without a code change if it's deleted for some reason.

Or did I miss something?

TomaszKolek: hm... yeah, probably.

budgetportal/models/pages.py
```diff
@@ -0,0 +1,145 @@
...
+    intro = RichTextField(blank=True)
+
+    content_panels = Page.content_panels + [FieldPanel("intro", classname="full")]
+
+    @route(r"^posts/")
+    def main(self, request):
+        from django.shortcuts import render
+
+        return render(request, self.get_template(request), self.get_context(request))
```
jbothma: I don't think you should have this

TomaszKolek: It was showed in one of tutorials I watched, trying all of those things (including printing inside)

budgetportal/settings.py
```diff
@@ -389,3 +408,5 @@
     "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.LimitOffsetPagination",
     "PAGE_SIZE": 20,
 }
+
+WAGTAIL_SITE_NAME = "My Wagtail Site"
```
jbothma: Reminder to make this `vulekamali` later

budgetportal/views.py
```diff
@@ -805,8 +805,7 @@ def resources(request):
 
 
 def guides(request, slug):
```
jbothma: we can delete this view once wagtail can take care of guides

requirements.txt
```diff
@@ -65,6 +65,7 @@ sqlparse==0.3.0
 tablib==1.1.0
 Unidecode==1.1.1
 urllib3==1.25.8
+wagtail==2.8.0
```
jbothma: please make this latest 2.7.x release - 2.8 just came out so who knows what bugs are there with the new features. 2.7.x is safer.

budgetportal/tests/helpers.py
```diff
@@ -45,3 +46,51 @@ def wait_until_text_in(self, selector, text):
         self.wait.until(
             EC.text_to_be_present_in_element((By.CSS_SELECTOR, selector), text)
         )
+
+    def _fixture_teardown(self):
```
jbothma: I want to come back to this later - not sure yet why this is needed. just hiding the file for now but keeping this comment

TomaszKolek: Once wagtail is installed there is some internal issue with TranscationalTaseCases.
The fix is to flush DB using allow cascade=True option.

We override _fixture_teardown to make it.

PR #274, opened 2020-03-13
Tomasz - Refactor

TomaszKolek:


budgetportal/views.py
```diff
@@ -821,8 +821,7 @@ def resources(request):
 
 
 def guides(request, slug):
-    if slug not in guide_data:
-        return HttpResponse(status=404)
+    slug = "index"
```
jbothma: this doesn't look like a good thing to merge at the moment

budgetportal/tests/helpers.py
```diff
@@ -50,3 +51,51 @@ def wait_until_text_in(self, selector, text):
         self.wait.until(
             EC.text_to_be_present_in_element((By.CSS_SELECTOR, selector), text)
         )
+
```
jbothma: do the tests pass without these changes at this stage?

budgetportal/views.py
TomaszKolek: true

PR #277, opened 2020-03-13
Add custom wagtail pages

TomaszKolek:


budgetportal/models.py
```diff
@@ -2083,3 +2087,137 @@ def csv_url(aggregate_url):
...
+    subpage_types = ["budgetportal.PostPage"]
+    max_count = 1
+    intro = RichTextField(blank=True)
+
+    content_panels = Page.content_panels + [FieldPanel("intro", classname="full")]
+
+
+class GuideIndexPage(Page):
+    max_count = 1
+    parent_page_types = ["budgetportal.LearningIndexPage"]
```
jbothma: no need to specify this in both directions - could just use one of parent_page_types or subpage_types

TomaszKolek: I'm not sure its a good aproach, maybe I'm doing something worng.

TomaszKolek: What I mean is its not working fully correctly without both.

TomaszKolek: As you cannot define root types. So under the root you can create all of pages without defining both

budgetportal/templates/base.html
```diff
@@ -10,18 +11,130 @@
...
+          https://app.appzi.io
+          https://d3js.org
+          {% comment %} Google Tag Manager HEAD inline {% endcomment %}
+          'sha256-v+9wGGBueHqSIuvMsUaFLaXTNw0D8igfDF+5k+9TVbE='
+        {% if debug %} {% comment %}for django-debug-toolbar{% endcomment %}
+          http://ajax.googleapis.com
+          'unsafe-eval'
+        {% endif %}
+        ;
+        frame-src *;
```
jbothma: reminder to self to check if we need to be this open

jbothma: userd to be
```
        'self'
        https://app.appzi.io
        https://www.youtube.com
        https://discussions.vulekamali.gov.za
        blob:;
```

budgetportal/models.py
jbothma: oh interesting...maybe change since v1. cool, doesn't hurt right now - just thought it could maintenance easier and reduce surprises

budgetportal/templates/base.html
```diff
@@ -10,18 +11,130 @@
...
+        img-src
```
jbothma:
```suggestion
        frame-src 'self'
                  https://app.appzi.io
                  https://www.youtube.com
                  https://discussions.vulekamali.gov.za
                  blob:;
        img-src
```

Please revert this - we don't need to relax this yet but we wanna merge this PR

budgetportal/templates/base.html
jbothma: reverting for now

PR #318, opened 2021-02-17
Fix appzi blocked by csp

Reagang:
Replaced old azure hostname with new one
Added the blocked script and frame source file too


requirements.txt
```diff
@@ -30,10 +30,10 @@ django-pipeline==1.7.0
 django-q==1.1.0
 django-recaptcha==2.0.6
 djangorestframework==3.11.0
-djangorestframework-csv==2.1.0
+djangorestframework-csv
 django-storages==1.9.1
 docopt==0.6.2
-drf-haystack==1.8.6
+drf-haystack
```
jbothma, line 36: why are you removing pinning dependencies? We pin them so that we don't get surprises when a different version is deployed that isn't compatible but isn't caught by automated tests.

PR #326, opened 2021-04-13
Indicate on department budget charts that they're budget and not actual 

Lunga001:


assets/scss/scenes/department/EconClassPackedCircles/econ-class-packed-circles.scss
```diff
@@ -36,6 +36,10 @@
       height: 100%;
       border: 0px;
     }
+
+    .marginTopMinus20 {
+      margin-top: -20px;
+    }
```
jbothma, line 42: it looks like you're mixing up "utility class" naming style with a component's specific style. utility classes should be defined outside of components, in the utility class stylesheets.

Since there's already some duplication going on here, I think it's fine to leave this in the component, but then we should name it for the specific component-part it relates to, e.g. see this suggestion.

Please apply the same change to the rest of the PR

```suggestion

    .legend {
      margin-top: -20px;
    }
```

jbothma: please see https://css-tricks.com/building-a-scalable-css-architecture-with-bem-and-utility-classes/ for utility classes and compare with https://css-tricks.com/developing-extensible-html-css-components/ for components

PR #349, opened 2022-11-30
eqprs file uploads

emre2038:


performance/models.py
```diff
@@ -7,18 +7,19 @@
...
-        f"eqprs_uploads/{uuid.uuid4()}/{filename}"
+        # f"eqprs_uploads/{uuid.uuid4()}/{filename}"
+        f"eqprs_uploads/{filename}"
     )
 
 
 class EQPRSFileUpload(models.Model):
-    user = models.ForeignKey(User, models.DO_NOTHING)
-    task = models.ForeignKey(Task, on_delete=models.CASCADE)
+    user = models.ForeignKey(User, models.DO_NOTHING, default=0)
```
jbothma, line 16: Why do we set a default here? Isnt it sufficient to  guarantee it gets set? Ideally setting it serverside so that user's can't mess with it at all like in the first example here https://books.agiliq.com/projects/django-admin-cookbook/en/latest/current_user.html

performance/models.py
```diff
@@ -7,18 +7,19 @@
 
 def eqprs_file_path(instance, filename):
     return (
-        f"eqprs_uploads/{uuid.uuid4()}/{filename}"
+        # f"eqprs_uploads/{uuid.uuid4()}/{filename}"
+        f"eqprs_uploads/{filename}"
```
jbothma, line 11: why are you removing the UUID? The point of the UUID is to ensure that the same filename doesn't replace an earlier upload's file

jbothma: perhaps we should document that

performance/models.py
```diff
@@ -7,18 +7,19 @@
...
     )
 
 
 class EQPRSFileUpload(models.Model):
-    user = models.ForeignKey(User, models.DO_NOTHING)
-    task = models.ForeignKey(Task, on_delete=models.CASCADE)
+    user = models.ForeignKey(User, models.DO_NOTHING, default=0)
+    task = models.ForeignKey(Task, on_delete=models.CASCADE, null=True, blank=True)
```
jbothma, line 17: why are you allowing blanks and nulls here? Can't we create the task on save and thus meet the db requirements?

performance/models.py
emre2038: this is to set the first user in the user dropdown as selected in default. happy to change if you think it's not the ideal way

performance/models.py
emre2038: I'm getting error with the UUID and I'll need your help.

performance/models.py
jbothma: mmm the 0 here refers to the user id in the database, not the position in the dropdown.
Also, this is a bit of a security issue - we don't want users to be able to set it as some other user, so it shouldn't happen client side - it should be set server-side. Please use the example above.

Note we want the creation user
Screenshot_2023-01-03_14-28-21

we **don't** want it to change every time it's edited:
Screenshot_2023-01-03_14-29-02

performance/tests/test_eqprs.py
```diff
@@ -0,0 +1,45 @@
...
+            is_active=True,
+        )
+        self.mocked_request = get_mocked_request(self.superuser)
+
+    def test_selected_user_is_current_user(self):
+        ea = EQPRSFileUploadAdmin(EQPRSFileUpload, AdminSite())
+        form = ea.get_form(self.mocked_request)
+        default_user = form.base_fields['user'].queryset.first()
+
+        self.assertEqual(default_user, self.superuser)
```
jbothma, line 37: this is something we want to validate as set on the object after save, not something set client-side.

jbothma: that is because we don't want users to be able to mess with the value that's set here

performance/models.py
```diff
@@ -35,11 +40,6 @@ class Frequency:
 
 
 class Indicator(models.Model):
-    """The indicator values available for a indicator in a department in a financial year"""
```
jbothma, line 38: you don't like my docs? :)

performance/models.py
jbothma: if we need to set _some_ value, I think changing to allowing null is better than setting it to zero - we never want to use an actual value when we mean "no value". 0 could be an actual user.

performance/admin.py
```diff
@@ -12,17 +171,49 @@ class EQPRSFileUploadAdmin(admin.ModelAdmin):
...
+            f = StringIO(full_text)
+            reader = csv.reader(f, delimiter=',')
+            parsed_data = [tuple(row) for row in reader]
+            parsed_data = [x for x in parsed_data if x]
+            del parsed_data[0:2]  # delete first 3 rows
+
+            frictionless_validated = validate_frictionless(parsed_data, obj.id)
+            if frictionless_validated:
+                del parsed_data[0]  # delete header row
+                threading.Thread(target=save_imported_indicators, args=(parsed_data, obj.id)).start()
```
jbothma, line 210: please use https://django-q.readthedocs.io/en/latest/tasks.html instead of starting a thread - I see where the misunderstanding came in :)

jbothma: this is why there's a task field in the upload model

performance/admin.py
```diff
@@ -12,17 +171,49 @@ class EQPRSFileUploadAdmin(admin.ModelAdmin):
...
+        return response
+
+    def save_model(self, request, obj, form, change):
+        super().save_model(request, obj, form, change)
+        full_text = obj.file.read().decode('utf-8')
+        report_type_validated = validate_report_type(full_text, obj.id)
+        if report_type_validated:
+            f = StringIO(full_text)
+            reader = csv.reader(f, delimiter=',')
+            parsed_data = [tuple(row) for row in reader]
```
jbothma, line 203: It's way more understandable and maintainable to use dictreader so that you can do key-based field lookup instead of index-based.

performance/admin.py
```diff
@@ -1,9 +1,168 @@
...
+        obj_to_update.num_imported = None
+        obj_to_update.num_not_imported = None
+        obj_to_update.import_report = generate_import_report(False, None, [])
+        obj_to_update.save()
+
+    return validated
+
+
+def validate_frictionless(data, obj_id):
+    report = validate(data)
```
jbothma, line 151: I think we should drop frictionless for the first iteration - especially if we switch to using key-based field lookup instead of index-based.

What I wanted to validate using frictionless was the presence of the required fields - see the suggestions below. But doing that in-memory on a server that might be doing this concurrently isn't obvious from the docs. I think instead of to_yaml in the docs links below can be swapped with to_dict. But I think we can waste a lot of time on this and I think we'll be safe enough with key-based lookup and key errors for a first iteration, and while we get the PR ready to ship, we can start the next iteration that will provide better user feedback using frictionless while we figure out the API.

For information:

I used the following schema to validate this excel file:

```json
{
  "fields": [
    {
      "type": "string",
      "name": "government",
      "constraints": {
        "pattern": "^([a-zA-Z-]+ )*[a-zA-Z-]+$"
      }
    },
    {
      "type": "string",
      "name": "group_id",
      "constraints": {
        "pattern": "^[a-z-]+$"
      }
    },
    {
      "type": "string",
      "name": "dataset_name"
    },
    {
      "type": "string",
      "name": "department_name"
    },
    {
      "type": "string",
      "name": "dataset_title"
    },
    {
      "type": "string",
      "name": "resource_name"
    },
    {
      "type": "string",
      "name": "resource_format"
    },
    {
      "type": "string",
      "name": "resource_url"
    }
  ]
}
```

[epre-budget-vote-docs-bulk-2022-06-22-upload.xlsx](https://github.com/vulekamali/datamanager/files/10337281/epre-budget-vote-docs-bulk-2022-06-22-upload.xlsx)
 with the command 

```bash
poetry run frictionless validate --schema-sync --schema budget-actual/bulk-budget-doc-metadata.json --limit-errors 3 budget-actual/epre-budget-vote-docs-bulk-2022-06-22-upload.xlsx
```

and I got the following output:

```
# -------
# invalid: budget-actual/epre-budget-vote-docs-bulk-2022-06-22-upload.xlsx 
# -------

## Summary 

The document was partially validated because of one of the limits
* limit errors 
* memory Limit

+-------------------------------------+-----------------------------------------------------------------+
| Description                         | Size/Name/Count                                                 |
+=====================================+=================================================================+
| File name (Not Found)               | budget-actual/epre-budget-vote-docs-bulk-2022-06-22-upload.xlsx |
+-------------------------------------+-----------------------------------------------------------------+
| File size                           | N/A                                                             |
+-------------------------------------+-----------------------------------------------------------------+
| Total Time Taken (sec)              | 0.136                                                           |
+-------------------------------------+-----------------------------------------------------------------+
| Rows Checked(Partial)**             | 72                                                              |
+-------------------------------------+-----------------------------------------------------------------+
| Total Errors                        | 3                                                               |
+-------------------------------------+-----------------------------------------------------------------+
| Constraint Error (constraint-error) | 3                                                               |
+-------------------------------------+-----------------------------------------------------------------+

## Errors 

+-------+---------+------------+----------------------------------------------------+
|   row |   field | code       | message                                            |
+=======+=========+============+====================================================+
|    70 |       1 | constraint | The cell "Mpumalanga " in row at position "70" and |
|       |         | -error     | field "government" at position "1" does not        |
|       |         |            | conform to a constraint: constraint "pattern" is   |
|       |         |            | "^([a-zA-Z-]+ )*[a-zA-Z-]+$"                       |
+-------+---------+------------+----------------------------------------------------+
|    71 |       2 | constraint | The cell "budget-vote_documents" in row at         |
|       |         | -error     | position "71" and field "group_id" at position "2" |
|       |         |            | does not conform to a constraint: constraint       |
|       |         |            | "pattern" is "^[a-z-]+$"                           |
+-------+---------+------------+----------------------------------------------------+
|    72 |       1 | constraint | The cell "Mpumalanga " in row at position "72" and |
|       |         | -error     | field "government" at position "1" does not        |
|       |         |            | conform to a constraint: constraint "pattern" is   |
|       |         |            | "^([a-zA-Z-]+ )*[a-zA-Z-]+$"                       |
+-------+---------+------------+----------------------------------------------------+
```

**only include the `name` fields** - the `type` and `constraint` fields are optional and we can skip them for now. 

The `--schema-sync` argument means "order the schema fields as per the data" so that column order doesn't break validation. I thin in the python method the argument is `validate(..., sync_schema=True)` https://github.com/frictionlessdata/framework/issues/380#issuecomment-698954928

https://v4.framework.frictionlessdata.io/docs/guides/validation-guide#validation-report

performance/admin.py
```diff
@@ -1,9 +1,168 @@
...
+                subtype=indicator_data[8],
+                mtsf_outcome=indicator_data[9],
+                cluster=indicator_data[10],
+                uid=indicator_data[43],
+            )
+            num_imported = num_imported + 1
+        elif department_name not in not_matching_departments:
+            not_matching_departments.append(department_name)
+
+    obj_to_update = models.EQPRSFileUpload.objects.get(id=obj_id)
```
jbothma, line 127: is this what you're trying to achieve here? https://docs.djangoproject.com/en/4.1/ref/models/instances/#django.db.models.Model.refresh_from_db

performance/static/js/eqprs-file-upload.js
```diff
@@ -0,0 +1,19 @@
...
+
+function formSubmit() {
+    enableDisableElement('id_user', false);
+}
+
+function enableDisableElement(id, enable) {
+    if (document.getElementById(id)) {
+        document.getElementById(id).disabled = enable;
+    }
+}
```
jbothma, line 19: I think this can be deleted when the creating user is set server-side

performance/admin.py
emre2038: done 👍

performance/models.py
emre2038: I don't remember why I removed it but probably I removed an extra line while removing `unique_together`

performance/models.py
emre2038: I wanted the user dropdown to be filled in default(even before it is saved). that's why I was playing with the `default` field. changed it now so the user is set in the backend

performance/models.py
emre2038: while creating the Task I need the object's id to set `source_id` of `Indicator`. So, first I need to save the object, then I can create the Task. That's why I allowed it to be null & blank.

performance/admin.py
emre2038: if I pass the object itself to the async task and try to update it there, I get connection already closed error, which makes sense. that's why I pass the id and find the object with id. I did this also because I thought this would make things easier in case we would call that async function from some other function in the future

performance/static/js/eqprs-file-upload.js
emre2038: done

performance/admin.py
emre2038: done

performance/admin.py
emre2038: removed validation

performance/tests/test_eqprs.py
emre2038: yea, I wanted to make sure the dropdown displayed the correct user as selected. now that we removed that dropdown, we don't need this test anymore

PR #351, opened 2022-12-13
Eqprs api

molotomt:


budgetportal/urls.py
```diff
@@ -267,6 +267,10 @@ def trigger_error(request):
     re_path(r"^", include(wagtail_urls)),
 ] + static.static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
+urlpatterns = [
```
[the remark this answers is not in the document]
molotomt: done

budgetportal/urls.py
```diff
@@ -248,6 +248,8 @@ def trigger_error(request):
     url(
         r"^robots\.txt$",
         views.robots,
+    ),    
+    path('', include('performance.urls')
```
jbothma, line 252: there's a duplicate one of these on row 270

performance/views.py
```diff
@@ -0,0 +1,15 @@
...
+from .serializer import IndicatorSerializer
+from rest_framework.viewsets import ModelViewSet
+from rest_framework import viewsets
+
+# Create your views here.
+
+class IndicatorReadOnlyModelViewSet(viewsets.ReadOnlyModelViewSet):
+    serializer_class = IndicatorSerializer
+    queryset = Indicator.objects.all()
+    http_method_names = ['get', 'put', 'head']
```
jbothma, line 12: to be readonly you need to disallow PUT as well

molotomt: thinking about it it make sense. done

performance/tests/test_eqprs_api.py
```diff
@@ -0,0 +1,35 @@
...
+# Create your tests here.
+
+
+class indicator_API_Test(APITestCase):
+    list_url = reverse("indicator-list")
+    fixture = ['test_api.json']
+    
+    def test_api(self):
+        response = self.client.get(self.list_url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
jbothma, line 18: I think a basic assertion that the API contains an expected field and value would be good. We don't need to test this comprehensively since we trust that django-rest-framework's model serializer is well-tested, but something rough like that there are as many objects in the response as expected, and that the indicator name is as we expect, would be good to detect if someone goes and totally breaks the django-rest-framework configuration

performance/views.py
```diff
@@ -1,3 +1,15 @@
 from django.shortcuts import render
+from .models import Indicator
+from .serializer import IndicatorSerializer
+from rest_framework.viewsets import ModelViewSet
+from rest_framework import viewsets
 
 # Create your views here.
```
jbothma, line 7: you can delete this now

performance/tests/test_eqprs_api.py
```diff
@@ -0,0 +1,35 @@
+#from django.test import TestCase
+import json
+from django.urls import reverse
+from rest_framework.test import APITestCase,APIRequestFactory
+from rest_framework import status
+from performance.views import IndicatorReadOnlyModelViewSet
+from performance.serializer import IndicatorSerializer
+from performance.models import Indicator
+# Create your tests here.
```
jbothma, line 9: you can delete this now

performance/tests/test_eqprs_api.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class indicator_API_Test(APITestCase):
+    list_url = reverse("indicator-list")
```
jbothma, line 13: what is this URL at the end of the day? I think it's worth hardcoding it here rather than using reverse() so that it isn't accidentally changed without noticing

performance/urls.py
```diff
@@ -0,0 +1,28 @@
+from django.contrib import admin
+from django.conf.urls import url
+from rest_framework.routers import DefaultRouter
+from django.urls import path, include
+from performance import views
+
+
+router = DefaultRouter()
+router.register('api/v1/eqprs', views.IndicatorReadOnlyModelViewSet)
```
jbothma, line 9:
```suggestion
router.register('eqprs', views.IndicatorReadOnlyModelViewSet)
```

budgetportal/urls.py
```diff
@@ -249,6 +249,7 @@ def trigger_error(request):
         r"^robots\.txt$",
         views.robots,
     ),
+    path("", include("performance.urls")),
```
jbothma, line 252:
```suggestion
    path("api/v1/performance", include("performance.urls")),
```

performance/urls.py
```diff
@@ -0,0 +1,28 @@
...
+#urlpatterns = router.urls
+
+urlpatterns =[
+    url('', include(router.urls))
```
jbothma, line 13:
```suggestion
    url('api/v1', include(router.urls))  
```

performance/urls.py
jbothma, line 9:
```suggestion
router.register('eqprs', views.IndicatorReadOnlyModelViewSet)
```

performance/tests/test_eqprs_api.py
jbothma: with my change suggestions in this PR , I think it will be something like `/performance/api/v1/eqprs`?

performance/tests/test_eqprs_api.py
jbothma: See how to access the response body at https://docs.djangoproject.com/en/2.2/topics/testing/tools/

performance/tests/test_eqprs_api.py
```diff
@@ -0,0 +1,56 @@
...
+        response_data=json.dumps(data) 
+        dict_json=json.loads(response_data) 
+       
+        found = False
+        message1 ="1.2.1 Percentage of valid invoices paid with in 30 days upon receipt by the department" 
+        message2 = "1.1.1 Unqualified audit opinion"
+        testvalue= ""
+        for data1 in dict_json:
+            for key, value in data1.items():
+                if key == "fields":
```
jbothma, line 29:
```python
data1["fields"]
```

performance/tests/test_eqprs_api.py
```diff
@@ -0,0 +1,56 @@
...
+                    for key2, value2 in value.items():
```
jbothma, line 30: you can just to `value["indicator_name"]` here - no need to loop

performance/tests/test_eqprs_api.py
```diff
@@ -0,0 +1,56 @@
...
+    list_url = '/api/v1/'
+    fixture = ['test_api.json']
+    
+    def test_api(self):
+        response = self.client.get(self.list_url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+           
+        data = json.load(open("performance/test_api.json"))
+        response_data=json.dumps(data) 
+        dict_json=json.loads(response_data)
```
jbothma, line 21: data and dict_json should be the same value - remember to delete your dev test code

performance/tests/test_eqprs_api.py
```diff
@@ -0,0 +1,56 @@
...
+
+        self.assertTrue(found, True)  
+               
+    def create(self):
+        response = self.client.post(self.list_url)
+        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)    
+
+    def update(self):
+        response = self.client.patch(self.list_url)
+        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
```
jbothma, line 47: I suspect these tests are not running because they're not named test_...

Try breaking them and see if you get failures before fixing.

```suggestion
    def test_create(self):
        response = self.client.post(self.list_url)
        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)    

    def test_update(self):
        response = self.client.patch(self.list_url)
        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)     
```

performance/tests/test_eqprs_api.py
```diff
@@ -1,56 +1,39 @@
...
-
-        #                     if message1 == testvalue or message2== testvalue:
-        #                         found = True
-
-        # self.assertTrue(found, True)
-
-    def create(self):
+        response = self.client.get(self.list_url)
+        self.assertEqual(response.status_code,200)
+        response = self.client.get(self.list_url).json()
```
jbothma, line 14: I wouldn't overwrite a variable that once had one type with another type - once I make response a HTTPResponse instance (or whatever), I wouldn't then replace it with the parsed body. I'd call it `response_payload` or `indicator_listing` or something (definitely not just `data` :) )

performance/tests/test_eqprs_api.py
```diff
@@ -1,56 +1,39 @@
...
+
+        found = False
+        notFound = False
+        message1 ="1.2.1 Percentage of valid invoices paid within 30 days upon receipt by the department"
+        message2 = "1.1.1 Unqualified audit opinion"
+        if response["results"][0]["indicator_name"] == message2 or response["results"][1]["indicator_name"] == message1:
+            found = True
+
+        if response["results"][0]["indicator_name"] != message2 or response["results"][1]["indicator_name"] != message1:
+            notFound = True
```
jbothma, line 24: I don't really understand the point of this - you can delete it I think. As well as the notFound assertion.

Remember Python variable names are snake_case not camelCase

performance/tests/test_eqprs_api.py
```diff
@@ -1,56 +1,39 @@
...
+
+        self.assertEqual(found, True)                
+        self.assertFalse(notFound, False)
+        self.assertTrue(found, True)
+        self.assertEqual(response["results"][0]["indicator_name"],message2)
+        self.assertEqual(response["results"][1]["indicator_name"],message1)
```
jbothma, line 30: remember the reason we're not doing this is that the ordering of the messages isn't guaranteed - that's why we're doing the loop or `or` approach as you've done above.

So I think you can delete this

PR #359, opened 2023-01-20
Eqprs api filtering v2

molotomt:


jbothma:
Switched to a different approach we could copy from another project.

PR #371, opened 2023-02-22
showcase updates

emre2038:


budgetportal/models/__init__.py
```diff
@@ -806,6 +806,7 @@ class ShowcaseItem(SortableMixin):
     cta_link_1 = models.URLField(
         null=True, blank=True, verbose_name="Call to action link 1"
     )
+    cta_enabled_2 = models.BooleanField(verbose_name="Enable call to action 2")
     cta_text_2 = models.CharField(max_length=200, verbose_name="Call to action text 2")
```
jbothma, line 810: I think a better UX for the admin user will be if it just doesn't show when it's blank.
We need both a) for the CTA label to be optional and b) for it not to be shown when we only want a primary.

I think using blank to decide not to show it is best. No need to support storing the value but not showing it. I think it would be more confusing and effort than helpful.
```suggestion
    cta_text_2 = models.CharField(max_length=200, verbose_name="Call to action text 2", blank=True)
```

emre2038: done

PR #372, opened 2023-02-23
name mapping

emre2038:


performance/tests/test_eqprs_uploads.py
```diff
@@ -153,3 +166,89 @@ def test_with_correct_csv(self):
...
+        )
+        performance.admin.save_imported_indicators(test_element.id)
+        test_element.refresh_from_db()
+
+        assert test_element.import_report == ""
+        assert test_element.num_imported == 1
+        indicator = models.Indicator.objects.all().first()
+        assert indicator.programme_name == "Programme 1: Administration"
+
+    def test_with_alias(self):
```
jbothma, line 233: please also add test coverage for

- if gov A and dept A is in the data, its old indicators for that financial year are deleted. Any other financial years remain
- if Gov A, B and dept A, B is in the data, old indicators for Gov A DeptA are deleted but Gov A dept B are not deleted

emre2038: done 👍

PR #376, opened 2023-03-06
adding and hiding columns

emre2038:


assets/js/components/performance/Table/scripts.jsx
```diff
@@ -73,17 +74,24 @@ class TabularView extends Component {
     }
 
     renderTableHead() {
```
jbothma, line 76: probably worth using a Set for this kind of things generally https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set

assets/js/components/performance/Table/scripts.jsx
```diff
@@ -279,40 +310,40 @@ class TabularView extends Component {
             <Grid container>
                 {this.renderSearchField()}
                 {this.renderFilter('financialYears',
-                                   'department__government__sphere__financial_year__slug',
-                                   'financialYears',
-                                   'Financial year',
-                                   'All financial years')}
+                    'department__government__sphere__financial_year__slug',
```
jbothma, line 313: this varied alignment of arguments (one at the end, the rest one indent level) is gonna be a bit hard to read, isn't it?

We traditionally follow airbnb formatting but I haven't run it anywhere in a long time, and I'm not sure if it's configured here.

How would it format this though? I think consistent is important, so perhaps the first arg should go onto a new line, if we don't want to indent the remaining args at the same level as the opening parenthesis?

Screenshot_2023-03-07_13-50-32

PR #381, opened 2023-03-13
title and description meta tags

emre2038:


performance/templates/performance/performance.html
```diff
@@ -1,7 +1,20 @@
...
+<meta property="og:title" content="{{ title }}" />
+{% endblock meta_title %}
+
+{% block meta_description %}
+{% with fallback_description="{{ description }}" %}
+<meta name="description" content="{{ description }}">
+<meta name="twitter:description" content="{{ description }}">
+<meta property="og:description" content="{{ description }}">
+{% endwith %}
+{% endblock meta_description %}
```
jbothma, line 15: merely setting the context variables should be enough https://github.com/vulekamali/datamanager/blob/master/budgetportal/templates/base.html#L110

emre2038: removed the tags @jbothma

PR #383, opened 2023-03-14
changed the truncation method of table cells

emre2038:


assets/js/components/performance/Table/scripts.jsx
```diff
@@ -308,6 +294,19 @@ class TabularView extends Component {
...
+            for (let entry of entries) {
+                entry.target.classList[entry.target.scrollHeight * 0.95 > entry.contentRect.height ? 'add' : 'remove']('truncated');
+            }
+        })
+
+        ps.forEach(p => {
+            observer.observe(p);
+        })
+    }
+
```
jbothma, line 309: is this a resource leak? how do we clean them up?
Looks like memory consumption is going up every time I click forward or back

jbothma: It looks like in prod it's garbage collecting. It might be because it's a prod js bundle - could you just look at garbage collection for the observers we're creating, and whether memory doesn't just grow infinitely as you use the UI?

Screenshot_2023-03-14_13-05-40

Screenshot_2023-03-14_12-55-17

PR #391, opened 2023-03-23
filtered table link

emre2038:


assets/js/components/performance/Table/scripts.jsx
```diff
@@ -98,6 +116,16 @@ class TabularView extends Component {
         let selectedFilters = this.state.selectedFilters;
         selectedFilters[name] = value;
 
+        let url = '';
+        Object.keys(selectedFilters)
+            .filter(key => selectedFilters[key] !== null)
+            .forEach((key, i) => {
+                const value = selectedFilters[key];
+                url += `${i === 0 ? '?' : '&'}${key}=${encodeURI(value)}`;
```
jbothma, line 124: I'd suggest using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString in future

PR #398, opened 2023-05-11
updated showcase item file input description

emre2038:


jbothma:
This will be super handy

Screenshot_2023-05-17_17-18-25

PR #421, opened 2023-08-24
Fix/infrastructure projects search action

paulmwatson:
Caused by `context` having `site.latest_year` but the HTML template only looking for `latest_year`. Added `site.latest_year` to the HTML template chain.


jbothma (approved):
nice find! thanks!

PR #425, opened 2023-09-05
iym data download

emre2038:


budgetportal/summaries.py
```diff
@@ -553,3 +557,39 @@ def get_aggregate_drilldowns(self):
...
+    def get_detail_csv_urls(self):
+        urls = {}
+        for fy in FinancialYear.get_available_years():
+            self.financial_year = fy.slug
+            aggr_url = self.get_detail_aggregate_url()
+            if aggr_url is not None:
+                url = csv_url(aggr_url)
+                urls[fy.slug] = url
+
+        return urls
```
jbothma, line 595: you're supposed to have one of these classes to produce the query for the in-year data.

One instance per dataset (therefore one instance per year)

budgetportal/summaries.py
```diff
@@ -553,3 +557,39 @@ def get_aggregate_drilldowns(self):
...
+        if openspending_api is not None:
+            department_ref = openspending_api.get_department_name_ref()
+            cuts = [
+                department_ref + ":" + "{}".format(department_name),
+            ]
+            aggregate_url = openspending_api.aggregate_url(cuts=cuts)
+
+            return aggregate_url
+        else:
+            return None
```
jbothma, line 584: you shouldn't need to override this. The implementation from the base class should be sufficient if you override get_aggregate_cuts and get_aggregate_drilldowns

budgetportal/summaries.py
emre2038: done

budgetportal/models/government.py
```diff
@@ -478,6 +477,24 @@ def get_estimates_of_econ_classes_expenditure_dataset(self, level=3):
...
+                    '+vocab_spheres: "' + sphere + '"',
+                    '+vocab_financial_years:"%s"' % financial_year,
+                    f"+name:national_in_year_spending_{financial_year}",
+                ]
+            )
+        }
+        response = ckan.action.package_search(**query)
+
+        if response["results"]:
+            return Dataset.from_package(response["results"][0])
```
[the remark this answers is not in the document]
emre2038: done

emre2038:
I updated `iym_datasets_json` function in views.py to use `BudgetedAndActualComparison` class @jbothma

assets/js/components/ChartSourceController/index.jsx
```diff
@@ -92,7 +92,8 @@ class ChartSourceController extends React.Component {
     }
 
     fetchActualExpenditureUrls(departmentName) {
-        let url = `../../actual-expenditure/?department_name=${encodeURI(departmentName)}`;
+        let selectedYear = document.querySelector('.YearSelect-item.is-active .YearSelect-link').innerText;
```
jbothma, line 95: don't use presentation stuff like page content inner text and css classes as data. Use data attributes if you want a reliable way to embed data.

There must be a variable in the page context which you can use for this.

budgetportal/datasets.py
```diff
@@ -369,3 +369,21 @@ def get_consolidated_expenditure_budget_dataset(financial_year):
         return Dataset.from_package(package)
     else:
         return None
+
+
+def get_budgeted_and_actual_comparison_dataset(financial_year):
```
jbothma, line 374: this is not  the budget and actual dataset, it's the in-year spending dataset.

```suggestion
def get_in_year_spending_dataset(financial_year):
```

budgetportal/summaries.py
```diff
@@ -553,3 +557,55 @@ def get_aggregate_drilldowns(self):
             openspending_api.get_programme_name_ref(),
             openspending_api.get_econ_class_4_ref(),
         ]
+
+
+class BudgetedAndActualComparison(DepartmentBudgetData):
```
jbothma, line 562:
```suggestion
class InYearSpending(DepartmentBudgetData):
```

This is not the Budgeted vs actual dataset, it's the In Year Spending dataset.

budgetportal/views.py
```diff
@@ -447,10 +448,24 @@ def department_page(
...
     context["eqprs_data_enabled"] = config.EQPRS_DATA_ENABLED
     context["in_year_spending_enabled"] = config.IN_YEAR_SPENDING_ENABLED
+    context["budgeted_and_actual_comparison"] = get_budgeted_and_actual_comparison_urls(
+        department, financial_years_context
+    )
 
     return render(request, "department.html", context)
 
 
+def get_budgeted_and_actual_comparison_urls(department, financial_years):
```
jbothma, line 458:
```suggestion
def get_in_year_spending_urls(department, financial_years):
```

budgetportal/summaries.py
```diff
@@ -553,3 +557,55 @@ def get_aggregate_drilldowns(self):
...
+    def __init__(self, department, financial_year):
+        super().__init__(department)
+        self.financial_year = financial_year
```
jbothma, line 565: any reason why you can't just use `department.government.sphere.financial_year` and avoid overriding this?

jbothma:
Some issues with the frontend


See the mockup in the trello card and here:


Let's also not show this bullet item at all if sphere is provincial or the feature flag is off (constance, django template variable).

Please add a line as part of the bullet item before the years row with the text

> In-year spending data:

If none of the years have data, please add the text

> No in-year spending data available for this department.

Please add another line after the year links as part of the bullet item with the text

> Note: In-year spending data download is in Thousands of Rands.

assets/js/components/ChartSourceController/index.jsx
emre2038: done

budgetportal/datasets.py
emre2038: done

budgetportal/summaries.py
emre2038: done

budgetportal/views.py
emre2038: done

budgetportal/summaries.py
emre2038: done

jbothma:
logging for later review

budgetportal/summaries.py
```diff
@@ -553,3 +557,53 @@ def get_aggregate_drilldowns(self):
...
+                openspending_api.get_programme_name_ref(),
+                openspending_api.get_department_name_ref(),
+                openspending_api.get_subprogramme_name_ref(),
+                openspending_api.get_econ_class_1_ref(),
+                openspending_api.get_econ_class_2_ref(),
+                openspending_api.get_econ_class_3_ref(),
+                openspending_api.get_econ_class_4_ref(),
+            ]
+        else:
+            return None
```
jbothma, line 591: we only want these drilldowns in the csv download. We don't want these drilldowns in the chart data. get_aggregate_url is the URL used in the chart, and get_detail_csv_url is the URL with more detail for download. Other charts use all drilldowns to get all the detail, but this dataset in this chart will have too much detail, which is why we should override get_detail_csv_url to use only these drilldowns.

jbothma: - don't override get_detail_csv_url - the csv download will use this via base class' get_detail_aggregate_url call
- return empty list of drilldowns in get_aggregate_drilldowns
- don't override get_aggregate_url - the chart will use this
- override get_detail_aggregate_url to use the custom list of drilldowns

emre2038: updated the code.

> return empty list of drilldowns in get_aggregate_drilldowns
instead of an empty list I had to return `None`. otherwise openspending returns an error.

I had to override `get_detail_csv_url` and `get_aggregate_url` to add an extra `None` check. Because in our case if the dataset does not exist in CKAN, `self.get_openspending_api()` returns `None`. The default functions throw error in these cases. But I also updated `get_detail_csv_url()` so it calls `get_detail_aggregate_url()` now instead of `get_aggregate_url()`.

Now the chart calls `get_aggregate_url()` function which has no drilldowns. Csv calls `get_detail_aggregate_url()` function which has a custom list of drilldowns

paulmwatson (approved):

PR #428, opened 2023-09-11
Testing CONN_MAX_AGE before attempting pooling

paulmwatson:


paulmwatson:
Setting `CONN_MAX_AGE` results in:
```
django.db.utils.OperationalError: database "test_budgetportal" is being accessed by other users
DETAIL:  There are 10 other sessions using the database.
```

